PR #10, opened 2023-10-26
Fix typo

trevyn:


Bromeon (commented):
Thanks! There's also another one: "catch-up" should be "catch up". Could you fix that as well, in the same commit?

trevyn:
Done!

Bromeon:
Thank you! 🙂

PR #9, opened 2023-10-19
Include more digestible bits for the introduction of the book

Wilker-uwu:
As discussed in the Discord server, the main problem with the tutorials and introductions included within the book is that it is hard to even grasp what even is an Extension in the context of Godot. Going back to read it a couple of times, the culprit for that is that it includes too much of the preparation and almost none of the process. Basically a lot of "You have to do that", "Do this before you begin using", "This is the difference compared to a version you might have never used", but too little of "This is how it works, here's how to set it up". 

While sure, assuming the reader might have the basic or or moderate understanding of the Rust language and its tools is reasonable to get the message delivered quickly, that does not exempt the book from the necessity of splitting the seemingly obvious details of the process to enable the explanation of its nuances, so keeping variations of "x is y+1 is 11" and having its variations as but a second thought will more often than not wrap it around into complicating it all again, both the learning process as well as the troubleshooting process.

This series of commits is aimed at easing the learning curve to new readers, both those new to Godot in general as well as those who may have just started learning Rust. The ideal scenario is that it will succeed in making a portion of the necessary information understandable at a glance for those who would prefer to scroll faster than reasonable. If all goes well, this should help other people to improve the explanations of the nuances that may affect the inclusion of the library into new and ongoing projects.

Thank you for reading :heart:

###### ps: yes it does have a diagram in it. sorry @Bromeon

PR #12, opened 2023-11-12
Added note regarding .obj file in some rust target builds causing God…

JamFactoryInc:
After struggling for a while with my own project, I found out that Godot was attempting to import an object binary file in my target dir with the extension '.obj' as a wavefront OBJ file, and was failing to build due to a supposed corrupted file. Adding a .gdignore fixed this issue, and I wanted to warn others in case they run into it.

### DIFF
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -98,7 +98,8 @@ so that it's easy to revert updates.
 ```admonish note
 It may be a advisable to put a [.gdignore][gd-ignore] file at the top-level of your rust project directory or within your `/target` dir.
 This is because for some Windows build targets, `cargo build` produces an object file with the `.obj` extension rather than the typical `.o`
-Godot will attempt to import this compiled binary as an Wavefront OBJ file (a common 3D model file, also using the extension `.obj`), and will fail to build due to a 'corrupted file' error.
+Godot will attempt to import this compiled binary as an Wavefront OBJ file (a common 3D model file, also using the extension `.obj`),
+and will fail to build due to a 'corrupted file' error.
 ```
 
 
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -101,6 +101,7 @@ This is because for some Windows build targets, `cargo build` produces an object
 Godot will attempt to import this compiled binary as an Wavefront OBJ file (a common 3D model file, also using the extension `.obj`), and will fail to build due to a 'corrupted file' error.
 ```
 
+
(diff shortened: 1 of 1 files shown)


src/intro/hello-world.md
```diff
@@ -95,6 +95,13 @@ The main crate of gdext is called `godot`. At this point, it is still hosted on
 To fetch the latest changes, you can regularly run a `cargo update` (possibly breaking). Keep your `Cargo.lock` file under version control,
 so that it's easy to revert updates.
 
+```admonish note
```
```suggestion
```admonish note title=".gdignore"
```

Bromeon (commented):
Thanks, great idea!
There are some conflicts after merging #9, could you resolve them?

src/intro/hello-world.md
```diff
@@ -95,6 +95,13 @@ The main crate of gdext is called `godot`. At this point, it is still hosted on
...
+It may be a advisable to put a [.gdignore][gd-ignore] file at the top-level of your rust project directory or within your `/target` dir.
+This is because for some Windows build targets, `cargo build` produces an object file with the `.obj` extension rather than the typical `.o`
```
PgBiel: I believe this message gives the slight impression that this is only recommended for Windows targets, when, in reality, this should be recommended for all platforms. Also, it should probably be made more explicit that this is only necessary if your Rust code is inside the Godot project folder (containing `project.godot`); having separate `godot/` and `rust/` folders would also solve the issue.

Bromeon: > having separate `godot/` and `rust/` folders would also solve the issue.

Btw, I just added that as the recommended directory structure in the Hello World tutorial 🙂 
https://godot-rust.github.io/book/intro/hello-world.html#directory-setup

PgBiel: Nice! Sounds like we could benefit from a direct link to that section then. Something like this:

"If you do not follow the [recommended directory structure setup for `gdext` projects](https://godot-rust.github.io/book/intro/hello-world.html#directory-setup) of having separate `godot/` and `rust/` folders, and instead place your Rust code root folder directly inside the Godot project folder (the one containing `project.godot`), then consider adding a `.gdignore` file at the root folder of your Rust code to avoid having the Godot editor accidentally and inappropriately attempt to import files from your Rust folder, such as `.obj` files which might be generated by `cargo build` and placed in your `target` subfolder, or other possibly conflicting files which the Rust compiler might generate."

Bromeon:
Thanks!

PR #14, opened 2023-11-26
Fix sprite2D.rotation() -> get_rotation() in hello-world.md

jamesmcm:
It seems the method is now named get_rotation() - tested with Godot 4.1 on Linux


Bromeon:
Thanks!

PR #16, opened 2023-11-30
Update `Gd::with_base` method recommendation to `Gd::from_init_fn`

Lamby777:
that thing I mentioned in the `#contrib-gdext` discord channel :P

Also added a reference at the bottom of the file (it used to be a broken reference)
image


Bromeon:
Thanks!

PR #21, opened 2024-01-05
Update `base` composition with `base()`, `base_mut()` methods.

Lemiczek:
Related to: https://github.com/godot-rust/gdext/pull/501

Reflects changes made by changing the way we access our `Base<T>`, essentially replacing our usage of `self.sprite` to instead utilize `base()` and `base_mut()`. 
There may be more places where information may be slightly inaccurate due to the [PR](https://github.com/godot-rust/gdext/pull/501). I looked over the book, and applied changes where it made sense. That said, I could've missed something, if so, I can amend. :+1:

### DIFF
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -275,8 +275,8 @@ Let's break this down.
 
 4. We define two fields `speed` and `angular_speed` for the logic. These are regular Rust fields, no magic involved. More about their use later.
 
-5. The `#[base]` attribute declares the `sprite` field, which allows `self` to access the base instance (via composition, as Rust does not have
-   native inheritance).
+5. The `#[base]` attribute declares the `sprite` field, which allows `self` to access the base instance (via `base()` or `base_mut()` as Rust does
+   not have native inheritance).
 
    - The field must have type `Base<T>`.
      - `T` must match the declared base class, e.g. `#[class(base=Sprite2D)]` implies `Base<Sprite2D>`.
@@ -342,7 +342,7 @@ impl ISprite2D for Player {
         // In GDScript, this would be: 
         // rotation += angular_speed * delta
         
-        self.sprite.rotate((self.angular_speed * delta) as f32);
+        self.base_mut().rotate((self.angular_speed * delta) as f32);
         // The 'rotate' method requires a f32, 
         // therefore we convert 'self.angular_speed * delta' which is a f64 to a f32
     }
@@ -351,6 +351,11 @@ impl ISprite2D for Player {
 
 GDScript uses property syntax here; Rust requires explicit method calls instead. Also, access to base class methods -- such as `rotate()`
 in this example -- is done via the `#[base]` field.
+
+```admonish warning
+  Make sure you are not using the `self.sprite` field directly. Use `base()` or `base_mut()` instead, otherwise you won't be able to access and call
+    the base class methods.
+```
 
 This is a point where you can compile your code, launch Godot and see the result. The sprite should rotate at a constant speed.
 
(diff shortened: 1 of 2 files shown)


Lemiczek:
Oops didn't lint, will fix :^)

src/intro/hello-world.md
```diff
@@ -275,8 +275,8 @@ Let's break this down.
 
 4. We define two fields `speed` and `angular_speed` for the logic. These are regular Rust fields, no magic involved. More about their use later.
 
-5. The `#[base]` attribute declares the `sprite` field, which allows `self` to access the base instance (via composition, as Rust does not have
-   native inheritance).
+5. The `#[base]` attribute declares the `sprite` field, which allows `self` to access the base instance (via `base()` or `base_mut()` as Rust does
+   not have native inheritance).
```
Bromeon: The old sentence is still true -- the point is that since no inheritance is possible, we use composition. The `base()/base_mut()` are just accessors.

It still makes sense to mention those, but it should be a new sentence and not replace the previous one.

src/intro/hello-world.md
```diff
@@ -383,15 +388,15 @@ impl ISprite2D for Player {
...
         let velocity = Vector2::UP.rotated(rotation) * self.speed as f32;
-        self.sprite.translate(velocity * delta as f32);
+        self.base_mut().translate(velocity * delta as f32);
         
         // or verbose: 
-        // self.sprite.set_position(
-        //     self.sprite.position() + velocity * delta as f32
+        // self.base_mut().set_position(
+        //     self.base().position() + velocity * delta as f32
         // );
```
Bromeon: I don't think this compiles -- there is a `&mut` borrow from `base_mut()` pending when `base()` is invoked.
Maybe change it to use two separate statements, storing the position in a variable.

Bromeon (commented):
Thank you! 🙂

src/intro/hello-world.md
phinjensen: I just ran into this while going through the book myself, and you're exactly right, it doesn't compile. Here's what I have, which does compile and functions correctly:

```rust
        let rotation = (self.angular_speed * delta) as f32;
        self.base_mut().rotate(rotation);

        let rotation = self.base().get_rotation();
        let velocity = Vector2::UP.rotated(rotation) * self.speed as f32;
        self.base_mut().translate(velocity * delta as f32)
```

Bromeon:
Sorry, I must merge this, there have already been 3 (!) people stumbling upon the outdated comment just in these few days. 
Implemented the remaining changes myself.

For the future, we should probably open book PRs alongside library PRs, or at least follow up within a day or so 🙂

daniellavoie:
Thank you very much for getting this through @bromeon ! 

A suggestion of mine to prevent such thing in the future would be to have the `gdext` and `book` projects versioned through release management with tags. If master is breakings stuff, it's okay, the docs website won't be impacted because instructions refers to versioned releases.

Happy to help on that contribution if you feel this is worthwhile.

Bromeon:
In this context, git tags seem like an extra maintenance burden without providing useful information to the user (as they don't follow SemVer). It's basically using the Git SHA with extra manual steps.

That said, things might be a bit clearer once the library is on crates.io, and we can let the book explicitly target a given version.
Before that, I don't really see need for action 😉

daniellavoie:
By git tags, I meant SemVer versioning that matches your crates.io release. (which I get that don't exist yet) All of these should allign. My suggestion was one for the bigger picture of overall release management.

Bromeon:
That definitely makes sense and is also what we have been doing [for gdnative](https://github.com/godot-rust/gdnative/tags) 👍

PR #25, opened 2024-01-09
fix borrow of angular_speed

good-praxis:
While working through the book, I got the following error:

```
error[E0503]: cannot use `self.angular_speed` because it was mutably borrowed
  --> src/player.rs:28:33
   |
28 |         self.base_mut().rotate((self.angular_speed * delta) as f32);
   |         ---------------         ^^^^^^^^^^^^^^^^^^                 - ... and the borrow might be used here, when that temporary is dropped and runs the destructor for type `BaseMut<'_, Player>`
   |         |                       |
   |         |                       use of borrowed `*self`
   |         `*self` is borrowed here
   |         a temporary with access to the borrow is created here ...
```

This commit changes the example to set a variable with the `rotate` argument first, I named that argument `radians` to match the parameter name.

### DIFF
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -343,7 +343,8 @@ impl ISprite2D for Player {
         // In GDScript, this would be: 
         // rotation += angular_speed * delta
         
-        self.base_mut().rotate((self.angular_speed * delta) as f32);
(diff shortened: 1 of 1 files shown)


Bromeon (approved):
Good catch, thanks a lot!

PR #31, opened 2024-01-29
singleton unregister fix + refcounted

StatisMike:


src/recipes/engine-singleton.md
```diff
@@ -81,16 +82,52 @@ unsafe impl ExtensionLibrary for MyExtension {
...
     fn on_level_deinit(level: InitLevel) {
         if level == InitLevel::Scene {
-            // Unregistering is needed to avoid memory leaks and 
-            // warnings, especially for hot reloading.
-            Engine::singleton().unregister_singleton(
-                StringName::from("MyEditorSingleton")
-            );
+            // Retrieve and keep the pointer to the `Engine` singleton, as
+            // we will need it for some operations.
+            let mut engine = Engine::singleton();
```
Bromeon: I would not even highlight "singleton" here, it distracts from the actual job at hand -- accessing a singleton.

Just mention
```suggestion
            // Get the Engine class instance.
            let mut engine = Engine::singleton();
```

src/recipes/engine-singleton.md
```diff
@@ -81,16 +82,52 @@ unsafe impl ExtensionLibrary for MyExtension {
...
+            // We need to retrieve the pointer to the singleton object,
+            // as it has to be freed manually - unregistering singleton 
+            // doesn't do it automatically.
+            let singleton = engine
+                .get_singleton(singleton_name.clone())
+                .expect("cannot retrieve the singleton");
+
+            // Unregistering singleton and freeing the object itself is needed 
+            // to avoid memory leaks and warnings, especially for hot reloading.
+            engine.unregister_singleton(singleton_name.clone());
```
Bromeon: No need to clone here.

src/recipes/engine-singleton.md
```diff
@@ -81,16 +82,52 @@ unsafe impl ExtensionLibrary for MyExtension {
...
 }
 ```
 
+```admonish warning
+As seen from above, the Godot engine, after singleton registration, isn't managing the singleton pointer. It has some implications for objects 
+whose class inherits from `RefCounted`, namely: Godot won't keep the object alive, which can lead to errors on its retrievals.
+
+Until `gdext` presents a more ergonomic API for singletons, the current solution used in the community is to wrap the pointer in a facade class 
+that itself doesn't inherit from `RefCounted`. Afterwards, the wrapper itself should be registered as a singleton.
+```
```
Bromeon: Please add a title to the warning.

First sentence could be a bit clearer:
> As seen above, the Godot engine does not manage the memory of the singleton pointer after registration.

Also, the second should be less vague than "some implications" -- I would link directly to https://github.com/godot-rust/gdext/issues/522 and straight discourage `RefCounted` objects from being used as singletons for the moment.

src/recipes/engine-singleton.md
```diff
@@ -81,16 +82,52 @@ unsafe impl ExtensionLibrary for MyExtension {
...
+}
+
+#[derive(GodotClass)]
+#[class(tool, init, base=Object)]
+struct MyEditorSingletonWrapper {
+    // Annotation allows exposing the RefCounted object on the GDScript side.
+    #[var]
+    inner: Gd<MyEditorSingleton>
+}
+```
```
Bromeon: Question: what's the benefit here? Why not simply let `MyEditorSingleton` inherit from `Object`?

Also, why the mix with editor classes (`#[tool]`)? I don't think it's relevant in this context.

Bromeon (changes requested):
Thanks a lot!
Added some comments. CI lints are also failing.

src/recipes/engine-singleton.md
StatisMike: `tool` was an omission - I've copied the declaration from the code block above. I guess both me and the initial author of this chapter were using singletons for some kind of editor tools, which was the source of the mistake. Will remove it also from the mentioned code block, as it isn't needed.

For this code block: I've intended this as a way to show the workaround if you **need** to make an object of a specific class a singleton, and this class happens to be inheriting from `RefCounted` (which happened to me during `gd-props` development). On second thought it may be a little too verbose for this chapter given that the circumstances are pretty specific. Hyperlink to the `gdext` issue you've mentioned above should be enough and will keep all of the information inside the warning block.

src/recipes/engine-singleton.md
StatisMike: Also removing the separate comment for `StringName` construction.

src/recipes/engine-singleton.md
```diff
@@ -81,16 +82,37 @@ unsafe impl ExtensionLibrary for MyExtension {
...
 ```
 
+```admonish warning title="Singletons inheriting from *RefCounted*"
+It is recommended to use manually-managed class as a base (often `Object` will be enough) for custom singletons. Godot engine isn't
+managing the instances registered as singletons on its own, so it's very possible that `RefCounted`-inheriting one will be dropped
+prematurely, leading to errors on its retrieval.
+
+If for any reason you need to have instance of reference-counted object registered as singleton, this [issue thread][refcounted-singleton-issue]
+presents some possible workarounds.
+```
```
Bromeon: This is not generally true; you can register singletons in GDScript just fine. It's likely a bug on our end or in the GDExtension API.

Please consider my previous comment:

> I would link directly to https://github.com/godot-rust/gdext/issues/522 and straight discourage `RefCounted` objects from being used as singletons for the moment.

Also, two general style issues:
- Your sentences miss a lot of articles ("a", "an", "the").
- Let's not speculate -- it's not "very possible" that `RefCounted` ones are dropped prematurely, this currently is _known_ to happen (due to a bug).

So, to keep this warning short, I'd simply higlight the facts **and remove everything else**.
- Only manually managed objects are currently supported, no classes inheriting from `RefCounted`.
- See issue #522 for more information (including workarounds).

Everything else is explained in the issue. If the information updates, there's one less place to keep in sync.

---

Sorry if this all sounds nitpicky, but it's important that we are a) concise and b) depict an accurate view of the status quo, including limitations.

Bromeon (commented):
Thanks for the update!

Bromeon:
Thanks! Lints are now failing because you used 1 character too many 😁 
Otherwise good to go! 👍

Bromeon (approved):

PR #32, opened 2024-02-01
Be more specific about the reader needs to do

jpaugh:
These changes reduce ambiguity, and give the reader a clearer picture of what they're supposed to do.

* Specify what kind of scene the reader needs to create
* Remove ambiguous reference to two project roots
* Reduce confusion about `res:///` paths when the user chooses a different folder for their `.gdextension`

### DIFF
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -44,7 +44,7 @@ project_dir
 We assume a Godot version of 4.1 or later. Feel free to download the latest stable one. You can download in-development ones,
 but we [do not provide official support for those][compatibility], so we recommend stable ones.
 
-Open the Godot project manager and create a new Godot 4 project in the `godot/` subfolder. Add a Sprite2D to the center of a new scene.
+Open the Godot project manager and create a new Godot 4 project in the `godot/` subfolder. Add a `Sprite2D` to the center of a new scene.
 We recommend that you follow the [Official tutorial][tutorial-begin] and stop at the point where it asks you to create a script.
 We will show you how to do that part in Rust.
 
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -44,8 +44,11 @@ project_dir
 We assume a Godot version of 4.1 or later. Feel free to download the latest stable one. You can download in-development ones,
 but we [do not provide official support for those][compatibility], so we recommend stable ones.
 
-Open the Godot project manager and create a blank Godot 4 project. Run the default scene to make sure everything is working.
-Save the changes and consider versioning each step of the tutorial in Git.
+Open the Godot project manager and create a new Godot 4 project in the godot/ subfolder. Add a Sprite2D to the center of a new scene.
+We recommend that you follow the [Official tutorial][tutorial-begin] and stop at the point where it asks you to create a script.
+We will show you how to do that part in Rust.
+
+Run your scene to make sure everything is working. Save the changes and consider versioning each step of the tutorial in Git.
 
 
 ## Create a Rust crate
@@ -279,6 +282,7 @@ Let's break this down.
 ```admonish warning title="Correct node type"
 When adding an instance of your `Player` class to the scene, make sure to select node type `Player` **and not its base `Sprite2D`**.
 Otherwise, your Rust logic will not run.
+We will guide you to make that change to your scene later, when you're ready to test it.
 
 If Godot fails to load a Rust class (e.g. due to an error in your extension), it may silently replace it with its base class.
 Use version control (git) to check for unwanted changes in `.tscn` files.
@@ -350,7 +354,12 @@ Do not use the `self.base` field directly. Use `self.base()` or `self.base_mut()
 the base class methods.
 ```
 
-This is a point where you can compile your code, launch Godot and see the result. The sprite should rotate at a constant speed.
+This is a point where you can see the result. Compile your code and launch the Godot editor.
+Right click on your Sprite2D in the scene tree, and choose "Change Type..."
+Find and choose the _Player_ node type, which will be a child of Sprite2D in the Change Type dialog that appears.
+
+Now, save your changes, and run the scene.
+The sprite should rotate at a constant speed.
 
 ![rotating sprite][img-sprite-rotating]
 
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -44,7 +44,7 @@ project_dir
 We assume a Godot version of 4.1 or later. Feel free to download the latest stable one. You can download in-development ones,
 but we [do not provide official support for those][compatibility], so we recommend stable ones.
 
-Open the Godot project manager and create a new Godot 4 project in the godot/ subfolder. Add a Sprite2D to the center of a new scene.
+Open the Godot project manager and create a new Godot 4 project in the `godot/` subfolder. Add a Sprite2D to the center of a new scene.
 We recommend that you follow the [Official tutorial][tutorial-begin] and stop at the point where it asks you to create a script.
 We will show you how to do that part in Rust.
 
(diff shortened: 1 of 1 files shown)


jpaugh:
There's one more change I'd like to make, but I don't quite understand what this text means. What's the difference between "outside" and "inside" in the following tip?

>  When exporting your project, you need to use paths inside `res://`.
> Outside paths like .. are not supported.
> [Source](https://godot-rust.github.io/book/intro/hello-world.html#admonition-export-paths)

This is especially confusing since the tutorial recommends using paths prefixed with `res://../rust/target/`, which is clearly _outside_ the godot resource root folder.

Would this be a better rewording?

> When exporting your project, you need to use `res://` paths. You cannot use filesystem paths.

src/intro/hello-world.md
```diff
@@ -125,8 +128,8 @@ $ ls -l
 This file tells Godot how to load your compiled Rust extension. It contains the path to the dynamic library, as well as the
 entry point (function) to initialize it with.
 
-First, add an empty `.gdextension` file anywhere in your project folder. In case you're familiar with Godot 3, this is the equivalent of
-`.gdnlib`. In this case, we create `res://HelloWorld.gdextension` in the root folder of the project and fill it as follows:
+First, add an empty `.gdextension` file anywhere in your `godot` subfolder. In case you're familiar with Godot 3, this is the equivalent of
+`.gdnlib`. In this case, we create `res://HelloWorld.gdextension` at `project_dir/godot` and fill it as follows:
```
Bromeon: > at `project_dir/godot` 

This is not very obvious, maybe something like "directly inside the `godot` directory"?

The term "project directory" is ambiguous (could mean the godot-rust project or the Godot project), so it might be better to avoid it here.

src/intro/hello-world.md
```diff
@@ -44,8 +44,11 @@ project_dir
 We assume a Godot version of 4.1 or later. Feel free to download the latest stable one. You can download in-development ones,
 but we [do not provide official support for those][compatibility], so we recommend stable ones.
 
-Open the Godot project manager and create a blank Godot 4 project. Run the default scene to make sure everything is working.
-Save the changes and consider versioning each step of the tutorial in Git.
+Open the Godot project manager and create a new Godot 4 project in the `godot/` subfolder. Add a Sprite2D to the center of a new scene.
+We recommend that you follow the [Official tutorial][tutorial-begin] and stop at the point where it asks you to create a script.
+We will show you how to do that part in Rust.
```
Bromeon: Add backticks around code/API symbols:
```suggestion
Open the Godot project manager and create a new Godot 4 project in the `godot/` subfolder. Add a `Sprite2D` to the center of a new scene.
We recommend that you follow the [Official tutorial][tutorial-begin] and stop at the point where it asks you to create a script.
We will show you how to do that part in Rust.
```

src/intro/hello-world.md
```diff
@@ -279,6 +282,7 @@ Let's break this down.
 ```admonish warning title="Correct node type"
 When adding an instance of your `Player` class to the scene, make sure to select node type `Player` **and not its base `Sprite2D`**.
 Otherwise, your Rust logic will not run.
+We will guide you to make that change to your scene later, when you're ready to test it.
```
Bromeon, line 284: This doesn't add much, I'd remove it.

We should keep warning boxes as short and to-the-point as possible, otherwise they become inflationary. Text inside boxes typically comes in the form of isolated explanations and not the typical tutorial flow (first we do this, later this...).

src/intro/hello-world.md
```diff
@@ -350,7 +354,12 @@ Do not use the `self.base` field directly. Use `self.base()` or `self.base_mut()
 the base class methods.
 ```
 
-This is a point where you can compile your code, launch Godot and see the result. The sprite should rotate at a constant speed.
+This is a point where you can see the result. Compile your code and launch the Godot editor.
+Right click on your Sprite2D in the scene tree, and choose "Change Type..."
+Find and choose the _Player_ node type, which will be a child of Sprite2D in the Change Type dialog that appears.
```
Bromeon: Here as well symbols in backticks (not italics) 🙂

src/intro/hello-world.md
```diff
@@ -350,7 +354,12 @@ Do not use the `self.base` field directly. Use `self.base()` or `self.base_mut()
...
+
+Now, save your changes, and run the scene.
+The sprite should rotate at a constant speed.
```
Bromeon: Could be on a single line; Markdown will combine it anyway.

Bromeon (commented):
Thanks a lot, these are great improvements! Added some comments, mostly minor formal things 😊 

---

> >  When exporting your project, you need to use paths inside `res://`.
> > Outside paths like .. are not supported.
>
> This is especially confusing since the tutorial recommends using paths prefixed with `res://../rust/target/`, which is clearly _outside_ the godot resource root folder.
>
> Would this be a better rewording?
>
> > When exporting your project, you need to use `res://` paths. You cannot use filesystem paths.

There is no contradiction because "outside paths" (`res://../something`) can be used during development, and this part talks about _exporting the project_.

I don't think your suggestion is clearer, because `res://../something` _is_ a `res://` path but can still not be used for exporting. The distinction of inside/outside is important, not filesystem or `res://`.

That said, I agree this could be clarified, but maybe it's enough to highlight the difference between development and exporting a bit better.

StatisMike:
> There is no contradiction because "outside paths" (`res://../something`) can be used during development, and this part talks about exporting the project.

@Bromeon To clarify: the `res:://.../` path seems to work with exported projects (at least on Linux, can't test it elsewhere). The library is kept outside the Godot binary and paths seems to be automatically adjusted.

Below project works both running from shell directly (`./gd-props\ tests.x86_64`) and through generated shell script.

Screenshot from 2024-02-01 17-20-47

Could be potentially useful if you make changes only in gdextension library and want to test changes on exported project, without needing to export whole Godot project on every iteration.

Bromeon:
@StatisMike Hm, interesting 🤔 but does that also work if you embed the `.pck` into the executable?

StatisMike:
@Bromeon Haven't tested it to that extent, haven't yet ventured to play with `.pck` files. Above was discovered by accident, by omission. I inteded to rework the test project structure to include `rust` dir inside `res://`. To my surprise - all worked without any changes from default `res://.../target/etc`, even on CI

jpaugh:
Okay, I'll get started on those changes this weekend.

> That said, I agree this could be clarified, but maybe it's enough to highlight the difference between development and exporting a bit better.

I definitely don't understand enough to fully ingest this explanation. What is "exporting" in this? Does it mean producing a final game binary?

Also, instead of clarifying, is it better to remove mention of topics that don't concern a newbie so much?

Bromeon:
> I definitely don't understand enough to fully ingest this explanation. What is "exporting" in this? Does it mean producing a final game binary?

https://docs.godotengine.org/en/stable/tutorials/export/exporting_projects.html

> 

> Also, instead of clarifying, is it better to remove mention of topics that don't concern a newbie so much?

Everyone has to go through exporting if they are ever shipping something (even if it's just a game jam on itch.io), and I think it's better to warn people early about paths, when it's still possible to refactor.

That said, StatisMike's observation changes things a bit. Maybe instead of telling people "outside paths cannot be exported", we could write something more neutral. You can copy/paste this verbatim:

```md
When working with paths (e.g. `res://` or `user://`), test whether they still work when 
[exporting a game][godot-export]. This is especially true for:

- Paths outside the project directory (e.g. `res://../rust/something`).
- Paths with incorrect case sensitivity -- those may work on Windows but not other targets.
``` 

[godot-export]: https://docs.godotengine.org/en/stable/tutorials/export/exporting_projects.html

src/intro/hello-world.md
jpaugh: Okay. I changed it to be consistent. Currently switching it to say "`godot` subfolder", since it also says that at line 47 when the folder structure is introduced. As for folder vs directory, folder is used 8 times and directory is used 5 times. Maybe making it consistent would be worthwhile. :shrug:

jpaugh:
@Bromeon I am finished with those changes you suggested. Thanks for the review.

Bromeon (approved):
Thanks a lot for your contribution! 🙂

PR #34, opened 2024-02-09
add required attribute to editor-plugin tutorial

vortexofdoom:
See title. Pasting the code directly from the tutorial will cause an error.

If we'd rather keep it to a self-contained `impl` block, I could change it to something like 
```rust 
#[godot_api]
impl MyEditorPlugin {}
```


Bromeon (approved):
Thanks!

PR #36, opened 2024-02-28
Update hello-world.md with note about adding code into the same file

tupoiu:
Hi!

I was going through Hello World in the godot-rust book, and my extension wasn't loading (I had put it into a separate file to lib.rs). It might be worth putting a small note in the hello world to stop other beginners like me making the same mistake!

### DIFF
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -244,7 +244,7 @@ Rust does not natively support inheritance, but the gdext API emulates it to a c
 
 ### Class declaration
 
-In this example, we declare a class called `Player`, which inherits [`Sprite2D`][api-sprite2d] (a node type):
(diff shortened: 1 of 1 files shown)


src/intro/hello-world.md
```diff
@@ -244,7 +244,7 @@ Rust does not natively support inheritance, but the gdext API emulates it to a c
 
 ### Class declaration
 
-In this example, we declare a class called `Player`, which inherits [`Sprite2D`][api-sprite2d] (a node type):
+In this example, we declare a class called `Player`, which inherits [`Sprite2D`][api-sprite2d] (a node type). Try adding this code into your `lib.rs` file:
```
Bromeon, line 247: This is incorrect -- classes can be defined anywhere, they don't need to be in `lib.rs`. 

See also `dodge-the-creeps` example. You may have encountered a different problem in your own code.

Bromeon:
Since the problem is likely elsewhere, I'll close this.
If you want to change something else, let me know and I'll glady reopen again 🙂

PR #43, opened 2024-04-29
Fix links in `recipes/custom-resources.rs`

oidro:
Fixes two links that previously redirected to the root directory of https://godot-rust.github.io/ instead of the `book` subdirectory.

(i.e. https://godot-rust.github.io/intro/hello-world.html -> https://godot-rust.github.io/book/intro/hello-world.html)

Hard to diagnose because it redirects just fine when viewing the book locally 😅


Bromeon:
Thank you!

PR #51, opened 2024-07-04
Update hello world example to use repo on crates.io

kyle-watson:
This is my first ever pull request, so please let me know if I messed anything up or if this isn't even a desired change. Also I am not positive if the Linux log is accurate as I compiled on windows and I replaced the parts of the log that looked similar.

### DIFF
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -71,19 +71,16 @@ edition = "2021"
 
 [lib]
 crate-type = ["cdylib"]  # Compile this crate to a dynamic C library.
-
-[dependencies]
-godot = { git = "https://github.com/godot-rust/gdext", branch = "master" }
 ```
 
 The `cdylib` crate type is not very common in Rust. Instead of building an application (`bin`) or a library to be utilized by other Rust code
 (`lib`), we create a _dynamic_ library, exposing an interface in the C programming language. This dynamic library is loaded by Godot at runtime,
 through the GDExtension interface.
 
-```admonish note title="Main crate"
-The main crate of gdext is called `godot`. At this point, it is still hosted on GitHub; in the future, it will be published to crates.io.
-To fetch the latest changes, you can regularly run a `cargo update` (possibly breaking). Keep your `Cargo.lock` file under version control, 
-so that it's easy to revert updates.
+Now add gdext to your project with:
+
+```bash
+cargo add godot
 ```
 
 To compile each iteration of the extension as you write code, you can use `cargo` as you normally do with any other Rust project:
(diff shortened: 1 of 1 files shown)


src/intro/hello-world.md
```diff
@@ -97,11 +91,11 @@ As an example, a Rust crate `hello` on Linux would be compiled to `libhello.so`:
...
-   Compiling godot4-prebuilt v0.0.0 
-       (https://github.com/godot-rust/godot4-prebuilt?branch=4.1.1#fca6897d)
-   Compiling proc-macro2 v1.0.69
-   [...]
-   Compiling godot v0.1.0 (https://github.com/godot-rust/gdext?branch=master#66df8f47)
+   Compiling godot-cell v0.1.1
+   Compiling glam v0.27.0
+   Compiling godot-core v0.1.1
+   Compiling godot-ffi v0.1.1
+   Compiling godot v0.1.1
```
Bromeon: I think this is a good example why we shouldn't include the log. It will already be outdated on next crates.io release again, and anytime we change dependencies, it becomes inaccurate. Furthermore, the output differs depending on the number of crates that have changed (it's not a full rebuild).

Could you remove the whole `cargo build` log?

The part about `ls -l` is also a bit strange, because `ls` doesn't typically show nested directories, nor does it put them in a frame. Maybe we should use a more conventional output here, following a `ls -l target/debug` command. But we can also just keep it OS-independent with a paragraph like

> After building, you should have the following files in the `target/debug` directory. Note that this applies to Linux, on Windows and macOS you will have different dynamic library names and file extensions.

src/intro/hello-world.md
```diff
@@ -73,19 +73,13 @@ edition = "2021"
 crate-type = ["cdylib"]  # Compile this crate to a dynamic C library.
 
 [dependencies]
-godot = { git = "https://github.com/godot-rust/gdext", branch = "master" }
+godot = "0.1"
 ```
```
Bromeon: I think we should keep both versions, for users who want a crates release and users who want to follow bleeding-edge development. Maybe mention the crates.io one first, and then introduce the GitHub one.

Bromeon (commented):
Thank you! It's a good change, we should probably use the opportunity to clean this up a bit, see comments.

Bromeon:
@kyle-watson Any update? 🙂

Bromeon:
Can we maybe recommend using `cargo add godot` instead of specifying the dependeny manually in `Cargo.toml`? Not only is that good practice, but it also prevents us from manually updating this page on every minor release...

src/intro/hello-world.md
```diff
@@ -91,7 +93,7 @@ This should output to `{YourCrate}/target/debug/` at least one variation of a co
 
 ```admonish tip
 For users who want to follow bleeding-edge development, you can directly link the github repo in your `Cargo.toml` dependencies section by replacing:
-`godot = "0.1"`
+`godot = "X.X.X"`
```
```suggestion
`godot = "0.x.y"`
```
So it's clear that the parts are different.

I think it's fair to update this if we get to version 1 😉

Bromeon:
@kyle-watson Do you have an update on this?

PR #53, opened 2024-08-06
Write a troubleshooting guide for initial setup

fpdotmonkey:
It's quite often that people come into the Discord because the hello-world isn't working and ask for help.  The issues they have usually are mundane configuration mistakes which are a bit tedious to help.

This addition will empower new users to solve their problems themselves and reduce load on the Q&A forum.  The advice provided is guided by experience from providing help many times in the past.


Bromeon:
This is a great idea, thanks a lot!

PR #56, opened 2024-10-01
updated instructions on exporting for web

illegalprime:
based on troubleshooting a recent project


src/toolchain/export-web.md
```diff
@@ -112,6 +113,7 @@ cargo +nightly build -Zbuild-std --target wasm32-unknown-emscripten
 
 Add a web export in the Godot Editor. In the top menu bar, go to `Project > Export...` and configure it there.
 Make sure to turn on the `Extensions Support` checkbox.
+In Godot 4.3 or above, make sure to turn on the `Thread Support` checkbox.
```
PgBiel, line 116: We do support builds without threads, but besides the Thread Support toggle in the editor, there are two steps:

1. Remove the `-pthread` flag (the line with `"-C", "link-args=-pthread"`) from `.cargo/config.toml` (it is necessary for thread support = on, and must be absent for thread support = off)
2. Enable the `experimental-wasm-nothreads` gdext feature flag.

Conversely, for Wasm binaries with thread support, you must have `-pthread` and compile _without_ the nothreads feature.

This means that Wasm binaries with and without thread support must be compiled separately. Then, you can add paths to both, allowing you to pick between the two by simply using the editor toggle, by modifying your `.gdextension` file like so (per https://github.com/godotengine/godot-cpp/pull/1451 ):

```toml
[libraries]
# ... other targets above ...
web.debug.threads.wasm32 = "res://bin/libgdexample.web.template_debug.wasm32.wasm"
web.release.threads.wasm32 = "res://bin/libgdexample.web.template_release.wasm32.wasm"
web.debug.wasm32 = "res://bin/libgdexample.web.template_debug.wasm32.nothreads.wasm"
web.release.wasm32 = "res://bin/libgdexample.web.template_release.wasm32.nothreads.wasm"
```

(Those paths may be changed to be more Rust-friendly, e.g. use `target/debug/whatever`. But note that, by default, Rust will produce the binaries at the exact same path regardless of the nothreads feature. We don't know yet how to solve that, so renaming will have to be manual.)

Then, once you compile both and place them in those paths within your project, toggling `Thread Support` on and off in the editor should work.

We should then mention both build modes in the manual, noting that builds without threads have better browser compatibility and don't require Cross-Origin Isolation support, while threaded builds have more limited compatibility and do require Cross-Origin Isolation support but may be more suitable if you depend on threading for your tasks.

src/toolchain/export-web.md
```diff
@@ -143,6 +145,12 @@ and open it in a supported browser such as Google Chrome or Microsoft Edge.
 
 [godot-export-templates]: https://docs.godotengine.org/en/stable/tutorials/export/exporting_projects.html#export-menu
 
+## Troubleshooting
+
+1. Make sure both `Extensions Support` and `Thread Support` (in Godot 4.3+) are turned on.
+2. Make sure the webserver you're running supports Cross-Origin Isolation (I used `sfz --coi`).
+3. Make sure your rust library and godot project are named differently, as sometimes `.wasm` files are overwritten.
+4. Make sure you're using the `emcc` version in the guide.
```
PgBiel, line 153: For `1` (regarding Thread Support), see other comment.

For `2`, it's worth mentioning that the Godot editor itself may be used to serve the web game for testing (there are images of this below), but you can also use its `serve.py` script directly if you want: https://github.com/godotengine/godot/blob/e3213aaef5e0e72b8272e65d989d3d8222be17ca/platform/web/serve.py

In addition, the main reason why support for non-threaded builds was added was precisely to work in environments without Cross-Origin Isolation IIRC, so that may be worth mentioning too.

Otherwise, good advice.

src/toolchain/export-web.md
```diff
@@ -155,9 +163,8 @@ for Chrome. It adds support for breakpoints and a memory viewer into the F12 men
 
 ---
 
-[^1]: Note: Due to a bug with `emscripten`, the maximum version of `emcc`[^2] that can one compile `Godot` with is `3.1.39`.  gdext itself should be able to support the latest version of `emcc`, however, it may be a safer bet to stick to version `3.1.39`.
+[^1]: Note: Due to a bug with `emscripten`, the maximum version of `emcc`[^2] that can one compile `Godot` with is `3.1.39`.  gdext itself should be able to support the latest version of `emcc`, however, it may be a safer bet to stick to version `3.1.39`. Anecdotally, `emcc-3.1.47` has also worked with `Godot 4.3`.
```
PgBiel, line 166: The info regarding the `emcc` version upper bound of `3.1.39` only applies to Godot 4.2 and below, and is thus now outdated. We should recommend 3.1.62 or later for Godot 4.3+. (Note that this is mentioned earlier in this document, so earlier mentions should also be updated.)

PgBiel (commented):
Thanks! Those are some good additions. Here are my initial comments (a bit large but only initially).

Bromeon:
@illegalprime are you still interested in pursuing this? If yes, could you apply @PgBiel's review feedback? Thanks 🙂

PgBiel:
FYI I'm currently busy with other things, but if there's no further interest in pushing this forward, I can try to pick this up in about a week or two. :)

Bromeon:
@PgBiel This would be absolutely awesome! Thanks a lot :heart:

zhouzhipeng:
hi @PgBiel i followed your comments and tried with 'thread support' off , but run with error : 
```
Uncaught (in promise) LinkError: WebAssembly.instantiate(): Import #28 "env" "lseek": imported function does not match the expected type 
```
 do u know why?

PgBiel:
Unfortunately errors related to WASM are most of the time rather inscrutable. Your best bet is to check everything you're doing to see if it's according to the instructions, including Godot and Emscripten versions, whether the flags used match the ones being used in this PR, and, if you're using thread support = off, make sure to remove the `-pthread` flag from `.cargo/config.toml` as well as enable the `experimental-wasm-nothreads` gdext feature. For further help, you can explain your situation along with your full setup in the gdext issue on WASM (https://github.com/godot-rust/gdext/issues/438).

Bromeon:
Superseded by #74.

PR #57, opened 2024-10-31
Update troubleshooting steps

AdminCrystal:
I personally got affected by this, contributing back to help others.

### DIFF
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -248,6 +248,8 @@ that should solve the most common problems.
   - The paths must also be relative to the directory that `project.godot` is in.  Typically it'll be `res://../rust/...`.
 - Have you written the Rust code necessary to generate the entry point symbol?
   - See [above](#rust-entry-point) for how.
+- Are you running matching versions of gdext and godot? By default gdext is up to date with latest godot release.
+  - You can specify a version to compile gdext with cargo features, format is api-x-x-x `godot = { ... , features = ["api-4-2-2"]}`
 - Is your directory structure like this below?  It's much easier when you ask for help if it is.
 
 ```txt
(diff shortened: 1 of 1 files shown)


Bromeon (commented):
Thanks a lot! 🙂 

I think the idea is good, however the way it is written conveys several misconceptions:
- The versions don't need to _match_. Godot >= gdext is required though.
- We don't keep the default up to date with the latest release, but with latest **minor** release.
- Recommending a specific API version like `api-4-2-2` is usually not a good idea, unless one needs APIs from 4.2.2 that aren't available in 4.2 already. But sticking to minor versions (`api-4-2`) increases the compatibility range without losing much.

This is all already explained in https://godot-rust.github.io/book/toolchain/godot-version.html, so I would not reiterate it at that point, but rather write something like:

> Are your gdext and Godot versions compatible? See [this page](...) for how to select the correct versions.

AdminCrystal:
> Thanks a lot! 🙂
> 
> I think the idea is good, however the way it is written conveys several misconceptions:
> 
> * The versions don't need to _match_. Godot >= gdext is required though.
> * We don't keep the default up to date with the latest release, but with latest **minor** release.
> * Recommending a specific API version like `api-4-2-2` is usually not a good idea, unless one needs APIs from 4.2.2 that aren't available in 4.2 already. But sticking to minor versions (`api-4-2`) increases the compatibility range without losing much.
> 
> This is all already explained in https://godot-rust.github.io/book/toolchain/godot-version.html, so I would not reiterate it at that point, but rather write something like:
> 
> > Are your gdext and Godot versions compatible? See [this page](...) for how to select the correct versions.

Thank you for explaining it that makes a lot of sense, have updated the pr to fit your comments and I agree linking to existing documentation is better than reiterating.

Bromeon:
Thanks! Renamed the label since `api-...` indicates a link to the API docs.

PR #58, opened 2024-11-06
Update dev-tools.md: add manually triggering CI

0x53A:
ref https://github.com/godot-rust/gdext/pull/933

~~Does hosting the picture on github work or should I check it directly into the repository?~~


0x53A:
Linter, bane of my existence

0x53A:
> Error: Error: Pushing to https://github.com/0x53A/book
remote: Permission to 0x53A/book.git denied to github-actions[bot].
fatal: unable to access 'https://github.com/0x53A/book/': The requested URL returned error: 403

If the PR itself is fine, i think you could copy my branch into your repository, close my PR and open a new one so the bot has permission to push the optimised image file,

Bromeon:
Thanks!

I've run into this issue before. I personally have the permissions to push commits, however the bot runs with a restricted setup. Our https://github.com/EndBug/add-and-commit action explains some stuff, but rather handwavy.

Experimented a bit with various permission and PATs, but I couldn't figure out how to allow the bot to push to _your_ branch (since it comes from your repo). I even tried triggering another workflow in the same repo (`workflow_dispatch` + `repository_dispatch`), but also doesn't seem possible with the default `GITHUB_TOKEN`. Security can really sometimes be annoying.

It _might_ work with `pull_request_target` instead of `pull_request`, but that's quite insecure and I couldn't checkout third-party code (from PRs), as that could be used to gain write access to the repo.

If you have any more ideas (other than manual invention), please let me know. I guess I could always just let `oxipng` fail and give the user instructions to fix the image themselves. Would just have been nice to automate this and not force contributors to go through the setup :confused:

0x53A:
> If you have any more ideas

instead of pushing directly to the branch of this PR (which the bot doesn't have permission), you could push to a branch of an additional fork, then send a PR against the _fork_. Then squash-merge the original PR to get rid of the history.

> I guess I could always just let oxipng fail and give the user instructions to fix the image themselves.

would be great to add instructions with the smallest distance to the error message, ideally in the exact same job:
image

0x53A:
well i did figure out how to run oxypng so finally green. At the least it's a rust tool so installation was painless.

0x53A:
Man I fucking love rust (and especially cargo) so much, it's unbelievable how much better it is than everything before it.

Today I was trying to program an STM microcontroller, and the C/C++ environment is just so, so bad, and rust+cargo+probe-rs just ... worked.

PR #59, opened 2024-11-06
add information on how to run oxipng

0x53A:


ReadMe.md
```diff
@@ -31,6 +31,13 @@ npm install --global markdownlint-cli2
 ```
 
 
+## Oxypng
+
+We use [oxpng](https://github.com/shssoichiro/oxipng) to optimize image file size.
+You can install it with ``cargo install oxipng``
+and then run it with ``oxipng --strip safe --alpha -r src``.
```
Bromeon:
```suggestion
## Oxipng

We use [oxipng](https://github.com/shssoichiro/oxipng) to optimize image file size.
You can install it with `cargo install oxipng` and then run it as follows:

```bash
oxipng --strip safe --alpha -r src
```
```

Comment (Bromeon) [reply ReadMe.md]: For some reason the \`\`\` cannot be nested, so the "suggestion" block doesn't render properly.

Either way, I think it's nice if the command has its own line, that makes it a bit easier to see and copy-paste 🙂

**Review by Bromeon — COMMENTED** [review]
Thanks! Note that it's called "oxipng", not "oxypng" 😉

File: ReadMe.md [hunk 2]
```diff
@@ -31,11 +31,13 @@ npm install --global markdownlint-cli2
...
+## Oxipng
 
-We use [oxpng](https://github.com/shssoichiro/oxipng) to optimize image file size.
-You can install it with ``cargo install oxipng``
-and then run it with ``oxipng --strip safe --alpha -r src``.
+We use [oxipng](https://github.com/shssoichiro/oxipng) to optimize image file size.
+You can install it with `cargo install oxipng` and then run it as follows:
+
+```bash
+oxipng --strip safe --alpha -r src
```
Bromeon, line 40: As mentioned, it wasn't rendered properly -- the bash codeblock still needs a closing triple backtick 🙂

0x53A: I did add closing ticks with another commit, or do you mean something else? image

Bromeon: You're right! For some reason this didn't display in my diff. Sorry!

Bromeon:
Thanks a lot :slightly_smiling_face:

PR #62, opened 2024-11-29
Update string argument passing documentation

PurpleStephyr:


Bromeon:
Thanks! There are quite a few other occurrences of `"string".into()` still in the book, would you mind updating them as well?

PurpleStephyr:
Ya, I can take a look. I am just starting with godot-rust and had noticed that this spot was wrong.

I think the "String types" section will likely need to be re-written - any suggestions for that? https://github.com/PurpleStephyr/godot-rust-book/blob/emit_signal_arg_passing/src/godot-api/builtins.md#string-types

Bromeon:
Oh, great catch!

Maybe this paragraph:

> These types all support `From` traits to convert to/from Rust `String`, and from `&str`. You can thus use `"some_string".into()` expressions.
> If you need more explicit typing, use `StringName::from("some_string")`.

Could be changed to something like the following. Feel free to quote reply and copy-paste 🙂 

---

When working with Godot APIs, you can pass references to the parameter type (e.g. `&GString`), as well as Rust strings `&str`, `&String`. To convert _between_ different string types in argument contexts (e.g. `StringName` -> `GString`), you can call `arg()`.

```rs
// Label::set_text() takes impl AsArg<GString>.
label.set_text("my text");
label.set_text(&string);           // Rust String
label.set_text(&gstring);          // GString
label.set_text(string_name.arg()); // StringName
```

Outside argument contexts, the `From` trait is implemented for conversions between strings: `GString::from("my text")` or `"my_text".into()`.

Bromeon:
I'll merge it as-is, other changes can be added separately. Thanks!

PR #63, opened 2024-11-29
Add missing rustflags for exporting to web

scroggo:
According to [a message in the Discord server](https://discord.com/channels/723850269347283004/1311869218572210207/1312024661860683847), you probably need some [more flags](https://discord.com/channels/723850269347283004/1274080437198651546/1275336695264575542) in order to properly export to the web. In my experience, these flags make it work! I have not tested to verify whether this is the exact set needed or a superset.

### DIFF
--- a/src/toolchain/export-web.md
+++ b/src/toolchain/export-web.md
@@ -74,7 +74,13 @@ rustflags = [
     "-C", "link-args=-sSIDE_MODULE=2",
     "-C", "link-args=-pthread", # was -sUSE_PTHREADS=1 in earlier emscripten versions
     "-C", "target-feature=+atomics,+bulk-memory,+mutable-globals",
-    "-Zlink-native-libraries=no"
+    "-Zlink-native-libraries=no",
+    "-Clink-arg=-fwasm-exceptions",
+    "-Clink-args=-sDISABLE_EXCEPTION_CATCHING=1",
+    "-Clink-args=-sEXPORT_ALL=1",
+    "-Clink-args=-sSUPPORT_LONGJMP=wasm",
+    "-Cllvm-args=-enable-emscripten-cxx-exceptions=0",
+    "-Cllvm-args=-wasm-enable-sjlj",
 ]
 ```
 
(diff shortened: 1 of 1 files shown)


PgBiel:
I believe the flag suggested in the existing PR should be enough, can you confirm? https://github.com/godot-rust/book/pull/56

scroggo:
> I believe the flag suggested in the existing PR should be enough, can you confirm? #56

@PgBiel , I'll list my steps here in case I did something wrong, but that flag by itself seemed to be insufficient for some features.

Steps:
- remove the new flags I added, except for `"-Cllvm-args=-enable-emscripten-cxx-exceptions=0"`
- run `cargo +nightly build -Zbuild-std --target wasm32-unknown-emscripten`
- From the Godot editor, click `Remote Debug` > `Run in Browser` [^1]

When the browser window opened, my game did work, **BUT** the sound did not[^2]. I don't know which of the other flags might be required to make sound work.

[^1]: Sometimes `Remote Debug` also includes options to `Re-export Project` and `Stop HTTPS Server`. I suspect that if they were showing I may have needed to click one or both of them in order to see the updated version?
[^2]: I restored the flags used here, rebuilt, stopped the server, and ran again, and the sound works again.

PgBiel:
> When the browser window opened, my game did work, BUT the sound did not

That's new. The singular flag in the existing PR should be enough to make it work with Godot 4.3+, but I hadn't heard about problems regarding sound. Would be nice if others could confirm that they can reproduce this problem first. (Maybe also post this information in https://github.com/godot-rust/gdext/issues/438 - that just that flag works but no sound, and the flags you list here work with sound as well - so others can see and try to reproduce)

scroggo:
In [this comment](https://github.com/godot-rust/gdext/issues/438#issuecomment-2513245928), @milesdesiran suggested 

> If @scroggo wants to add the book documentation like in [godot-rust/book#56 ](https://github.com/godot-rust/book/pull/56) , that would be helpful - that PR has similar changes, but has been left unattended by the creator for a few months.

We can pull more from that PR into this one, though it looks like there's still some discussion about some of the particulars and I'm not sure what best to recommend. I think adding this line (now that I have removed some lines that turned out to not be necessary) would be beneficial by itself.

PgBiel:
> I think adding this line (now that I have removed some lines that turned out to not be necessary) would be beneficial by itself.

I tend to agree, even though further improvements to this docs page are planned, I think there isn't harm in just adding this flag for now and potentially helping out some users.

cc @Bromeon for further thoughts - if you agree, I think this PR could be merged as is.

Bromeon:
Thanks a lot! :+1:

PR #66, opened 2024-12-17
Add a point in the Hello World troubleshooting docs

lenny991:
Added a point which I missed from earlier in the documentation into the troubleshooting part of hello world. I think it should be in troubleshooting as well, as it would've helped me when starting to work with this addon.

### DIFF
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -236,7 +236,7 @@ that should solve the most common problems.
   - See [above](#rust-entry-point) for how.
 - Are your gdext and Godot versions compatible? See [this page][versioning] for how to select the correct versions.
 - Is your directory structure like this below?  It's much easier when you ask for help if it is.
-- Do you have godot in your `PATH` as `godot4` or an environment variable called `GODOT4_BIN`, containing the path to the Godot executable?
+- Do you have Godot in your `PATH` as `godot4` or an environment variable called `GODOT4_BIN`, containing the path to the Godot executable?
 
 ```txt
 my-cool-project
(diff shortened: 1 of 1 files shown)


Bromeon:
Thanks a lot! This only applies to the `api-custom` feature though, so maybe it should be prefixed with

> In case you use `api-custom`, ...

---

Also, would be good to keep the bullet point with the file structure next to the actual file structure :slightly_smiling_face:

lenny991:
Hey, could you elaborate on this api-custom thing? I was not aware I was using it.

lenny991:
Hey, could you elaborate on this api-custom thing? I was not aware I was using it.

Bromeon:
Basically this:
https://godot-rust.github.io/book/toolchain/godot-version.html#custom-godot-versions

If you weren't using it, I wonder why you need `godot4` or `GODOT4_BIN`? Are you sure those are necessary?

lenny991:
Adding the $GODOT_BIN env variable fixed my problem of the crate nothing working (meaning the whole program just halted when communicating from GDScript to rust). Can it get turned on automatically? Can this have something to do with the fact I'm on Linux? Of course it's possible the problem lied somewhere else but as far as I remember that's the only thing I changed.

Bromeon:
This sounds suspicious.

We access the `GODOT4_BIN` variable [here](https://github.com/godot-rust/gdext/blob/3c56b3fed129d6ba38dd47173a7b024d78c8b522/godot-bindings/src/godot_exe.rs#L261-L262). And this function is invoked in 3 places, all of which are directly or indirectly gated by `#[cfg(feature = "api-custom")]`.

So either you have this feature enabled, or it's something else.

---

So I don't want to add a blanket "set these env vars" recommendation, because that's not the truth and propagates the myth that those are necessary at all times :wink: when it's worded conditionally, that's already much better.

lenny991:
Yes of course, now it's worded better and in a better place. It seems to be so that my earlier issue was from somewhere else which is unfortunate, as this would've been a nice and easy fix which I could've shared to others. In any case it's working now and I think this is a worthwhile addition to the docs.

Bromeon:
Thanks a lot!

PR #71, opened 2025-01-18
Added line to remind you to include the player file in your lib

Relu-Avali:
After having spent an unreasonable amount of time trying to figure out why my node was not showing up in the editor I feel a line like this would have been a big time saver.

### DIFF
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -264,7 +264,9 @@ Rust does not natively support inheritance, but the gdext API emulates it to a c
 
 ### Class declaration
 
-In this example, we declare a class called `Player`, which inherits [`Sprite2D`][api-class-sprite2d] (a node type):
+In this example, we declare a class called `Player`, which inherits Sprite2D (a node type).
+This can be either defined in `lib.rs` or in a separate file `player.rs`.
+In case you go for the latter, don't forget to declare mod player; in your lib.rs file.
 
 ```rust
 use godot::prelude::*;
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -266,7 +266,7 @@ Rust does not natively support inheritance, but the gdext API emulates it to a c
 
 In this example, we declare a class called `Player`, which inherits Sprite2D (a node type).
 This can be either defined in `lib.rs` or in a separate file `player.rs`.
-In case you go for the latter, don't forget to declare mod player; in your lib.rs file.
+In case you go for the latter, don't forget to declare `mod player;` in your `lib.rs` file.
 
 ```rust
 use godot::prelude::*;
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -264,7 +264,7 @@ Rust does not natively support inheritance, but the gdext API emulates it to a c
 
 ### Class declaration
 
-In this example, we declare a class called `Player`, which inherits Sprite2D (a node type).
+In this example, we declare a class called `Player`, which inherits `Sprite2D` (a node type).
 This can be either defined in `lib.rs` or in a separate file `player.rs`.
 In case you go for the latter, don't forget to declare `mod player;` in your `lib.rs` file.
 


Bromeon:
Hey, thank you! :slightly_smiling_face: 

However, the page currently doesn't mention you should write your code in `player.rs`, so this was an assumption of yours? Maybe we should be explicit that users can use either `lib.rs` or a dedicated `player.rs` file, but the sentence 

> Add the line mod player; to your lib.rs file. 

on its own adds more questions than it answers.

Relu-Avali:
> Hey, thank you! 🙂
> 
> However, the page currently doesn't mention you should write your code in `player.rs`, so this was an assumption of yours? Maybe we should be explicit that users can use either `lib.rs` or a dedicated `player.rs` file, but the sentence
> 
> > Add the line mod player; to your lib.rs file.
> 
> on its own adds more questions than it answers.

You are correct, I may have jumped too fast on this one. Out of habit I placed all player code on its own `player.rs` file, to me it was a very natural thing to do. 

I would like to pitch it as an idea/suggestion in that case:
Have the documentations make the reader create a `player.rs` file and move all player code into it. I think makes more sense having it in its own file instead of `lib.rs`. It would be better clean code practice and also teach the reader that for any node to show in the Godot editor it must be included in the compiled Rust create.

Bromeon:
> I would like to pitch it as an idea/suggestion in that case:
> Have the documentations make the reader create a `player.rs` file and move all player code into it. I think makes more sense having it in its own file instead of `lib.rs`. It would be better clean code practice and also teach the reader that for any node to show in the Godot editor it must be included in the compiled Rust create.

Since this is a "Hello World" example, we should keep the number of introduced concepts to a minimum, and focus on the important parts to distract too much -- i.e. on how the entry point and classes are defined.

Knowing how Rust modules work [is a prerequesite of the book](https://godot-rust.github.io/book/intro/index.html#admonition-note), but I see that it may be worth mentioning just to avoid the confusion you encountered. But it shouldn't be more than a side-note, and it shouldn't end up in code -- users are perfectly within their rights to use `lib.rs` for everything, especially during Hello World.

As such, I'd extend the existing paragraph in [_Class declaration_](https://godot-rust.github.io/book/intro/hello-world.html#class-declaration):

> In this example, we declare a class called `Player`, which inherits [Sprite2D](https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.Sprite2D.html) (a node type).
> This can be either defined in `lib.rs` or in a separate file `player.rs`. In case you go for the latter, don't forget to declare `mod player;` in your `lib.rs` file.

What do you think?

Relu-Avali:
> >
> Since this is a "Hello World" example, we should keep the number of introduced concepts to a minimum, and focus on the important parts to distract too much -- i.e. on how the entry point and classes are defined.
> 
> Knowing how Rust modules work [is a prerequesite of the book](https://godot-rust.github.io/book/intro/index.html#admonition-note), but I see that it may be worth mentioning just to avoid the confusion you encountered. But it shouldn't be more than a side-note, and it shouldn't end up in code -- users are perfectly within their rights to use `lib.rs` for everything, especially during Hello World.
> 
> As such, I'd extend the existing paragraph in [_Class declaration_](https://godot-rust.github.io/book/intro/hello-world.html#class-declaration):
> 
> > In this example, we declare a class called `Player`, which inherits [Sprite2D](https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.Sprite2D.html) (a node type).
> > This can be either defined in `lib.rs` or in a separate file `player.rs`. In case you go for the latter, don't forget to declare `mod player;` in your `lib.rs` file.
> 
> What do you think?

I agree that number of new concepts should be kept to a minimum in a hello world example. This all stems from a mistake I made. Am I the first one to create a new file for player code and not include it on the create? If I am the only one I do not think any changes are necessary. If this a reoccurring mistake I agree that your suggested changes are good.

Bromeon:
Hard to say if this occurs often, but even if not, mentioning it briefly in a sentence (like I suggested) may clear things up.

My point was mostly to not dedicate too much content to "how to structure modules in Rust", because that's a topic that people should already know. So showing how to _apply_ that knowledge is good, but _teaching_ it is a bit too much :slightly_smiling_face:

Bromeon:
So, what do you think about [my suggestion](https://github.com/godot-rust/book/pull/71#issuecomment-2600919616) earlier?

> As such, I'd extend the existing paragraph in [_Class declaration_](https://godot-rust.github.io/book/intro/hello-world.html#class-declaration):
> 
> > In this example, we declare a class called `Player`, which inherits [Sprite2D](https://godot-rust.github.io/docs/gdext/master/godot/classes/struct.Sprite2D.html) (a node type).
> > This can be either defined in `lib.rs` or in a separate file `player.rs`. In case you go for the latter, don't forget to declare `mod player;` in your `lib.rs` file.
> 
> What do you think?

Relu-Avali:
> > What do you think?

I think it is a good suggestion, I like it.

Bromeon:
I meant, do you want to update your PR to reflect this? :wink:

Relu-Avali:
I have updated the PR based on previous discussion.

src/intro/hello-world.md
```diff
@@ -264,7 +264,9 @@ Rust does not natively support inheritance, but the gdext API emulates it to a c
 
 ### Class declaration
 
-In this example, we declare a class called `Player`, which inherits [`Sprite2D`][api-class-sprite2d] (a node type):
+In this example, we declare a class called `Player`, which inherits Sprite2D (a node type).
+This can be either defined in `lib.rs` or in a separate file `player.rs`.
+In case you go for the latter, don't forget to declare mod player; in your lib.rs file.
```
```suggestion
In case you go for the latter, don't forget to declare `mod player;` in your `lib.rs` file.
```

src/intro/hello-world.md
```diff
@@ -264,7 +264,9 @@ Rust does not natively support inheritance, but the gdext API emulates it to a c
 
 ### Class declaration
 
-In this example, we declare a class called `Player`, which inherits [`Sprite2D`][api-class-sprite2d] (a node type):
+In this example, we declare a class called `Player`, which inherits Sprite2D (a node type).
```
```suggestion
In this example, we declare a class called `Player`, which inherits `Sprite2D` (a node type).
```

Bromeon:
Thank you!

PR #72, opened 2025-01-31
Minor typo in prelude description

reshen:
I think this is a bit clearer since we're talking about the crate in the preceding example code as opposed to the library as a whole.

### DIFF
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -282,7 +282,7 @@ struct Player {
 
 Let's break this down.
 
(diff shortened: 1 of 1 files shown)


Bromeon:
You might need to put it in backticks, so it's seen as a symbol and doesn't trigger CI lint:
```
`godot`
```

reshen:
Thanks Bromeon, should be fixed now.

Bromeon:
Thank you!

PR #73, opened 2025-02-02
Replaces deprecated #[init(default = ...)] with #[init(val = ...)]

ColinWttt:
The attribute key #[init(val = ...)] replaces #[init(default = ...)]. More information on https://github.com/godot-rust/gdext/pull/844


Bromeon:
Great catch, thank you!

PR #79, opened 2025-02-28
Use `::class_name()` instead of string literal 

greenfox1505:
IMHO, for portability, it should be suggested you put your name in your struct, instead of placing the string inside the register/de-register method.

### DIFF
--- a/src/recipes/engine-singleton.md
+++ b/src/recipes/engine-singleton.md
@@ -54,6 +54,7 @@ struct MyEditorSingleton {
 impl MyEditorSingleton {
     #[func]
     fn foo(&mut self) {}
+    const NAME:&str = "MyEditorSingleton";
 }
 ```
 
@@ -74,7 +75,7 @@ unsafe impl ExtensionLibrary for MyExtension {
             // The `&str` identifies your singleton and can be
             // used later to access it.
             Engine::singleton().register_singleton(
-                "MyEngineSingleton",
+                MyEngineSingleton::NAME,
                 &MyEngineSingleton::new_alloc(),
             );
         }
@@ -85,15 +86,14 @@ unsafe impl ExtensionLibrary for MyExtension {
             // Let's keep a variable of our Engine singleton instance,
             // and MyEngineSingleton name.
             let mut engine = Engine::singleton();
-            let singleton_name = "MyEngineSingleton";
 
             // Here, we manually retrieve our singleton(s) that we've registered,
             // so we can unregister them and free them from memory - unregistering
             // singletons isn't handled automatically by the library.
-            if let Some(my_singleton) = engine.get_singleton(singleton_name) {
+            if let Some(my_singleton) = engine.get_singleton(MyEngineSingleton::NAME) {
                 // Unregistering from Godot, and freeing from memory is required
                 // to avoid memory leaks, warnings, and hot reloading problems.
-                engine.unregister_singleton(singleton_name);
+                engine.unregister_singleton(MyEngineSingleton::NAME);
                 my_singleton.free();
             } else {
                 // You can either recover, or panic from here.
--- a/src/recipes/engine-singleton.md
+++ b/src/recipes/engine-singleton.md
@@ -54,7 +54,6 @@ struct MyEditorSingleton {
 impl MyEditorSingleton {
     #[func]
     fn foo(&mut self) {}
-    const NAME:&str = "MyEditorSingleton";
 }
 ```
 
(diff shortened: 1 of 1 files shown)


Bromeon:
If we already do that, then probably `class_name()` is better than a user-maintained constant?

greenfox1505:
I thought about suggesting something like: `std::any::type_name::<Self>().into()` but then if you're gunna do that, you might as well put it in the trait. Which turns into scope creep. I just wanted a standard place to look from names, but if y'all wanna turn it into something better, please do.

Yarwin:
While we are at it, I usually use something like that for user-defined singletons:

```rs
pub trait UserSingleton:
    GodotClass + Bounds<Declarer = bounds::DeclUser> + NewAlloc + Inherits<Object>
{
    const NAME: &'static str;

    fn singleton() -> Gd<Self> {
        Engine::singleton()
            .get_singleton(Self::NAME)
            .unwrap()
            .cast::<Self>()
    }

    fn initialize() -> Gd<Self> {
        let game_system = Self::new_alloc();
        Engine::singleton().register_singleton(Self::NAME, &game_system);
        game_system
    }

    fn exit(&mut self) {
        Engine::singleton().unregister_singleton(Self::NAME);
    }

    #[allow(unused_variables)]
    fn physics_process(&mut self, delta: f64) {}

    #[allow(unused_variables)]
    fn process(&mut self, delta: f64) {}
}
```

It allows to ensure that user-defined singletons can be used in the same fashion as engine-declared/code generated ones

Bromeon:
Imo these suggestions, while good ones, miss the point of that page: how to declare a singleton in the easiest way.

Fancy abstractions -- and singleton registration itself -- should eventually be provided in the library.

So what about this? It already exists:


> If we already do that, then probably `class_name()` is better than a user-maintained constant?

Yarwin:
>  If we already do that, then probably class_name() is better than a user-maintained constant?

Note: I tested it briefly with gdscript and seems to work fine :+1: – I haven't noticed any clashes nor issues with namespace. 

<details>
<summary> Example code </summary>

```rs
struct MyExtension;


#[gdextension]
unsafe impl ExtensionLibrary for MyExtension {
    fn on_level_init(level: InitLevel) {
        if level == InitLevel::Scene {
            MyEditorSingleton::initialize();
        }
    }

    fn on_level_deinit(level: InitLevel) {
        if level == InitLevel::Scene {
            let mut my_singleton = MyEditorSingleton::singleton();
            my_singleton.bind_mut().exit();
            my_singleton.free();
        }
    }
}

#[derive(GodotClass)]
#[class(init, base=Object)]
struct MyEditorSingleton {
    base: Base<Object>,
}

#[godot_api]
impl MyEditorSingleton {
    #[func]
    fn foo(&mut self) {
        godot_print!("foo");
    }
}

impl UserSingleton for MyEditorSingleton {} // everything is provided by default impl - no macro magic needed 

pub trait UserSingleton:
GodotClass + Bounds<Declarer = bounds::DeclUser> + NewAlloc + Inherits<Object>
{

    fn singleton() -> Gd<Self> {
        Engine::singleton()
            .get_singleton(&Self::class_name().to_string_name())
            .unwrap()
            .cast::<Self>()
    }

    fn initialize() -> Gd<Self> {
        let game_system = Self::new_alloc();
        Engine::singleton().register_singleton(&Self::class_name().to_string_name(), &game_system);
        game_system
    }

    fn exit(&mut self) {
        Engine::singleton().unregister_singleton(&Self::class_name().to_string_name());
    }

    #[allow(unused_variables)]
    fn physics_process(&mut self, delta: f64) {} // in real world implementation should be handled inside the class; docs should provide ways to call it (via the GameLoop class)

    #[allow(unused_variables)]
    fn process(&mut self, delta: f64) {} // as above
}
```

</details>

image

Bromeon:
Thanks!

(Maybe we should at some point add `impl AsArg for ClassName`...)

Bromeon:
@greenfox1505 could you change this to use `::class_name()`, which avoids defining a new constant?

I'll then consider `impl AsArg<StringName> for ClassName` in parallel.

greenfox1505:
I know it's been a while, but it sounded like this workflow was gunna change so I wasn't gunna worry about it.

I ended up needing this docs again so I found myself doing the same "fix" to the example, so I figured I'd update this, now with `::class_name()`.

In the future, I think we could register `#[class(tool,base=Object,init,singleton)]`?

Bromeon:
Thanks for the update!

> In the future, I think we could register `#[class(tool,base=Object,init,singleton)]`?

Yes, see https://github.com/godot-rust/gdext/issues/1060 :slightly_smiling_face:

PR #81, opened 2025-03-03
Fix: replace outdated info

ColinWttt:
fix: double `init`

### DIFF
--- a/src/recipes/custom-resources.md
+++ b/src/recipes/custom-resources.md
@@ -25,7 +25,7 @@ struct ResourceType {
 
 The above resource does not export any variables. While not all resources require exported variables, most do.
 
-If your custom resource has lifecycle methods that need to run in the editor,
+To execute code within the Godot Editor (e.g., overriding `set_property` to run custom logic when editing Inspector properties),
 you should annotate the class with `#[class(tool)]`.
 
 ```rust
--- a/src/recipes/custom-resources.md
+++ b/src/recipes/custom-resources.md
@@ -25,7 +25,7 @@ struct ResourceType {
 
 The above resource does not export any variables. While not all resources require exported variables, most do.
 
-If your custom resource has lifecycle methods (`ready()`, `process()` etc.) that need to run in the editor,
+If your custom resource has lifecycle methods that need to run in the editor,
 you should annotate the class with `#[class(tool)]`.
 
 ```rust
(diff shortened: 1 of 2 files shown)


src/recipes/custom-resources.md
```diff
@@ -37,7 +37,8 @@ struct ResourceType {
 
 #[godot_api]
 impl IResource for ResourceType {
-  fn init(base: Base<Resource>) -> Self { ... }
+    fn ready(&mut self) { ... }
+    fn process(&mut self, delta: f64) { ... }
```
Yarwin: Why this change? Resource does not implement these virtual methods :thinking: 

https://godot-rust.github.io/docs/gdext/master/godot/classes/trait.IResource.html

Yarwin:
hey hey

Thanks you for suggesting change in `compatibility.md`! I think it is valuable info. I don't get the change for resource's recipe though :thinking:

src/recipes/custom-resources.md
ColinWttt: You’re right, I made a mistake. When should a resource have `#[class(tool)]`, considering it doesn’t have `ready` or `process`?

Yarwin: In case when it runs in the editor. Docs outline few uses: https://docs.godotengine.org/en/stable/tutorials/plugins/running_code_in_the_editor.html

src/recipes/custom-resources.md
```diff
@@ -25,7 +25,7 @@ struct ResourceType {
 
 The above resource does not export any variables. While not all resources require exported variables, most do.
 
-If your custom resource has lifecycle methods (`ready()`, `process()` etc.) that need to run in the editor,
+If your custom resource has lifecycle methods that need to run in the editor,
```
Bromeon: Good call to correct this :+1: 

But now I think it's no longer clear what is meant, as "lifecycle methods" isn't such a well-known term -- and the `Resource` class doesn't even have those.

Is there anything concrete we could mention here?

ColinWttt: I used a simple example. The `#[class(tool)]` has many uses, and I’m not sure whether to list them all here. I also added all `impl IResource` methods below.

src/recipes/custom-resources.md
```diff
@@ -37,7 +37,30 @@ struct ResourceType {
...
+    fn get_property_list(&mut self) -> Vec<PropertyInfo> { ... }
+
+    // Called whenever Godot retrieves value of property. Allows to customize existing properties.
+    // Every property info goes through this method, except properties added with get_property_list().
+    fn validate_property(&self, property: &mut PropertyInfo) { ... }
+
+    // Called by Godot to tell if a property has a custom revert or not.
+    fn property_get_revert(&self, property: StringName) -> Option<Variant> { ... }
+
+    fn setup_local_to_scene(&mut self) { ... }
```
Bromeon, line 63: Could you add a brief comment here too? Then it should be ready to merge :slightly_smiling_face:

ColinWttt: These comments are copy and paste from [IResource docs](https://godot-rust.github.io/docs/gdext/master/godot/classes/trait.IResource.html), and gdext does not have comments for this method. I found the [corresponding explanation in Godot](https://docs.godotengine.org/zh-cn/4.x/classes/class_resource.html#class-resource-private-method-setup-local-to-scene), but I prefer to add it after gdext includes its official comments, as there may be inconsistencies.

Bromeon:
Thanks!

PR #88, opened 2025-03-30
Adjust target feature instructions for Wasm threading

PgBiel:
Based on findings from https://github.com/godot-rust/gdext/pull/1105 and https://github.com/godot-rust/gdext/pull/1107, I propose:

1. Suggesting removing `-Ctarget-features=+atomics` when compiling a nothreads build, to ensure `std` is compiled without multi-threading (using polyfills for threading APIs).
2. Suggesting not adding the `+mutable-globals` target feature anymore in any wasm builds, as it appears to be stable and enabled by default now, per https://doc.rust-lang.org/1.85.1/rustc/platform-support/wasm32-unknown-unknown.html#enabled-webassembly-features
3. **Tentatively** suggesting removing the `+bulk-memory` target feature as well. In principle, some error messages from emscripten suggest to me that `+atomics` is sufficient here, as removing this flag made my rather simple test project  (with async signals for fancier testing) work all the same. However, let's keep an eye on this in case someone has a problem with this.
4. Updating some minor things

### DIFF
--- a/src/toolchain/export-web.md
+++ b/src/toolchain/export-web.md
@@ -362,7 +362,7 @@ as otherwise your extension's `.wasm` file may be overwritten, leading to confus
     likely not help at all.
 
     Some common panic causes include:
-      - Attempting to call certain threaded code in a `nothreads` build, such as `std::thread::spawn(...)` or `thread_local!`;
+      - Attempting to call certain multi-threaded code in a `nothreads` build, such as `std::thread::spawn(...)`;
       - Using panicking variants of methods, such as `Array::at` instead of `Array::get`;
       - Calling `.unwrap()` on `Option::None` or `Result::Err`.
 
--- a/src/toolchain/export-web.md
+++ b/src/toolchain/export-web.md
@@ -77,8 +77,8 @@ Start by adding the following contents to that file:
 [target.wasm32-unknown-emscripten]
 rustflags = [
     "-C", "link-args=-pthread", # /!\ Read 'Thread support' below regarding this flag
+    "-C", "target-feature=+atomics", # /!\ Read 'Thread support' below regarding this flag
     "-C", "link-args=-sSIDE_MODULE=2",
-    "-C", "target-feature=+atomics,+bulk-memory,+mutable-globals",
     "-Zlink-native-libraries=no",
     "-Cllvm-args=-enable-emscripten-cxx-exceptions=0",
 ]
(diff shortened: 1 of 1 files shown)


src/toolchain/export-web.md
```diff
@@ -133,16 +133,22 @@ With the proposed initial configuration from ["Project configuration"](#project-
 your extension would break. If you'd like your extension to support builds without multi-threading as well to avoid this problem,
 you will need to update your build setup in one of the two following ways.
 
-Note that earlier versions of emscripten expected `link-args=-sUSE_PTHREADS=1` instead of `link-args=-pthread`, but this flag has
+```admonish note
+Ensure you're using recommended and recent versions of emscripten (3.1.x) and nightly Rust (at least Rust 1.85 is recommended).
```
Bromeon: Emscripten is already at version 4, "recent" may be misleading here. Especially if 4.x is not mentioned, this sounds like already outdated advice.

Weren't there some specific versions (or a range) that was deemed "good" in our last discussion? I recently tested with `emcc 3.1.62`... 

If the context is specifically only the deprecated flags, then maybe we should just recommend a minimum lower bound (without the "recent" qualifier).

src/toolchain/export-web.md
```diff
@@ -445,4 +450,4 @@ to compile your extension as well, but newer `emcc` versions might still work re
 
 [^2]: `emcc` is the name of Emscripten's compiler.
 
-[^3]: The primary reason for this is it is necessary to compile with `-sSHARED_MEMORY` enabled. The shipped `std` does not, so building `std` is a requirement. Related info on about WASM support can be found [here](https://github.com/rust-lang/rust/issues/77839).
+[^3]: The primary reason for this is it is necessary to compile with `-pthread` and `-Ctarget-feature=+atomics` enabled for multi-threaded builds. The shipped `std` does not, and may also build with other flags we don't use, so building `std` is a requirement. Related info on about WASM support can be found [here](https://github.com/rust-lang/rust/issues/77839).
```
Bromeon: Could you split this onto multiple lines?

Bromeon (commented):
Thanks a lot! Good to know that the usage of flags gets simpler with more recent emcc/rustc versions :+1:

src/toolchain/export-web.md
PgBiel: You're right. I've made the message more explicit.

src/toolchain/export-web.md
PgBiel: Sure thing

Bromeon:
Thanks!

PR #89, opened 2025-03-30
Add "delete cache" to troubleshoot section in hello-world.md

### LINKED ISSUE #1102: Hot Reload of [export] does not work
I just tried gdext for the first time. I pretty much built exactly what is described in https://godot-rust.github.io/book/intro/hello-world.html , only with an additional exported property.

```rust
struct Player {
    #[export]
    speed: i32,
```

My understanding is that, since I am using Godot 4.4, hot reloading of my code should work and a simple `cargo build` should cause the "speed" property to appear in the editor on my Player node. But it just doesn't - only when I reload the entire project does the property show up. Notable, it does show up, so the compiled binaries seem to be connected correctly, it's just that hot reloading doesn't work.

Are there any additional restrictions that need to be satisfied to enable that feature?

atollk:
Closes https://github.com/godot-rust/gdext/issues/1102

### DIFF
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -235,6 +235,7 @@ that should solve the most common problems.
 - Have you written the Rust code necessary to generate the entry point symbol?
   - See [above](#rust-entry-point) for how.
 - Are your gdext and Godot versions compatible? See [this page][versioning] for how to select the correct versions.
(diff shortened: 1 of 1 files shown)


Bromeon:
Thank you!

PR #93, opened 2025-04-15
Add `termdot` to Ecosystem.

Joezeo:


Joezeo:
Seems the license-guard check was cancelled, what should I do🧐

Yarwin:
I updated ubuntu runner for license guard  #94, it should be all good after rebasing on current master

Bromeon:
Thanks a lot!

Regarding category, is this not an editor plugin? In your [ReadMe](https://github.com/termdot/termdot?tab=readme-ov-file#installation), you write:
> In your scene, add a Termdot node. This is the main control node for the plugin.

Joezeo:
It's an external terminal, but needs Godot plugin support

Bromeon:
Then it would still fit under _Editor Plugins > Project workflow_.
_Applications_ are more standalone programs built with Godot, but not necessarily integrating into the editor.

Could you move it? :slightly_smiling_face:

Bromeon:
Thanks!

PR #98, opened 2025-05-07
Fix typo in Android export documentation

Zaprit:
Small PR to fix typo in NDK, and capitalise JDK and NDK as they are both acronyms.


Yarwin:
Looks good to me

PR #102, opened 2025-07-03
ecosystem: Add Stag Toolkit

arocull:
Bromeon suggested the idea to include this like 6 months ago, ahaha. Opening a PR, but if it doesn't fit then that's chill.

This plugin is a hybrid of gdext and GDScript. I wasn't sure what category my plugin falls under since it's general-purpose, but current Rust-dependent features of the plugin include:
- `IslandBuilder` node - Real-time meshing and collision generation for level whiteboxing
- `SimulatedRope`, `SimulatedRopeBinding` nodes - Interactive 3D rope simulation 
- `QueueFloat` object - Statistical analysis tool for large numbers

Stuff I might do in the future may be animations, physics, and/or data-vis related.

Thanks!


Yarwin:
Looks good to me + I'm familiar with this toolkit already

PR #103, opened 2025-07-09
Added clearification that multiple godot_api blocks may be needed for signals

Relu-Avali:


src/register/signals.md
```diff
@@ -335,9 +335,10 @@ This still allows you to use `#[signal]` and will register each signal declared
 The typed signal API is generated for your class under the following conditions:
 
 - Your class declares a `Base<T>` field.
-- You have a `#[godot_api]` block (empty if necessary).
-   - This is a technical limitation that may be lifted in the future.
 - You do not opt out from typed signals with `no_typed_signals`.
+- You have a `#[godot_api]` block  on your class `impl` (empty if necessary).
+   - Requires seperate blocks for both `impl Monster` and `impl INode3D for Monster`
+   - This is a technical limitation that may be lifted in the future.
```
Bromeon: It only requires the attribute on the inherent `impl` block, not the trait one -- or am I missing something?

If yes, you can combine this to 1 point:

```suggestion
- You have a `#[godot_api]` block  on your class' inherent `impl` block (not the `I*` trait `impl`).
  Leave it empty if necessary. This is a technical limitation that may be lifted in the future.
```

Bromeon (commented):
Thank you!

src/register/signals.md
Relu-Avali: The reason for this PR is that I was struggling for a while with the following code:

```rust
#[derive(GodotClass)]
#[class(base=Node)]
pub struct Monster {
    base: Base<Node>,
}

#[godot_api]
impl INode for Monster { 
    fn init(base: Base<Node>) -> Self {
        Self {
            base,
        }
    }
}

impl Monster {
  pub fn set_signals(monster: Gd<Monster>) {
    //Does not work, trait bound not satisfied
    monster.signals().tree_exiting().connect(Monster::on_tree_exiting);
    
    //Works when first cast to base node
    let upcast_monster = monster.upcast::<Node>();
    upcast_monster.signals().tree_exiting().connect(Monster::on_tree_exiting);
  }
  
  fn on_tree_exiting() {
      godot_print!("Monster is exiting!");
  }
}
```

The docs currently does not make it clear that an `godot_api` block is also needed on the `impl Monster` in order for it to work.

Bromeon: I see, but how does that change my suggestion?

The point is that `#[godot_api]`  is required for the inherent `impl`.

Relu-Avali: When you write "inherent" you are referring to the rust standard block and not the one inherited from the godot node?

edit: I was not aware those were two different words with different meaning. Yes, your suggestion makes sense now

Bromeon: Yes exactly, [inherent `impl`](https://doc.rust-lang.org/reference/items/implementations.html#inherent-implementations) is the one that doesn't implement a trait.

PR #105, opened 2025-07-23
Fix Resource saver and loader example code

beicause:
- Fixes typo and incomplete `on_level_deinit()`
- Adds unregister Resource loader and saver when the extension exits.

### DIFF
--- a/src/recipes/resource-saver-loader.md
+++ b/src/recipes/resource-saver-loader.md
@@ -31,16 +31,19 @@ unsafe impl ExtensionLibrary for MyGDExtension {
     fn on_level_init(level: InitLevel) {
         if level == InitLevel::Scene {
             Engine::singleton().register_singleton(
-                "MyAssetSingleton",
+                &MyAssetSingleton::class_name().to_string_name(),
                 &MyAssetSingleton::new_alloc(),
             );
         }
     }
 
-    // Unregisters the singleton when the extension exits.
-    fn on_level_init(level: InitLevel) {
+    // Unregister the singleton when the extension is unloaded.
+    fn on_level_deinit(level: InitLevel) {
         if level == InitLevel::Scene {
-            Engine::singleton().unregister_singleton("MyAssetSingleton");
+            let mut engine = Engine::singleton();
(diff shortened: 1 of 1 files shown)


src/recipes/resource-saver-loader.md
```diff
@@ -37,11 +37,17 @@ unsafe impl ExtensionLibrary for MyGDExtension {
...
-            Engine::singleton().unregister_singleton("MyAssetSingleton");
-            my_singleton.free();
+            let mut engine = Engine::singleton();
+            let singleton_name = "MyAssetSingleton";
+            if let Some(my_singleton) = engine.get_singleton(singleton_name) {
+                engine.unregister_singleton(singleton_name);
+                my_singleton.free();
+            } else {
+                godot_error!("Failed to get singleton");
+            }
```
Bromeon: When exactly can this fail? After all, we registered it during init.

Generally I'd rather not encourage defensive programming if it's not necessary, it creates uncertainty about the reliability of operations. If there's a valid reason why this may fail, it should be explicitly documented.

src/recipes/resource-saver-loader.md
```diff
@@ -81,6 +87,14 @@ impl IObject for MyAssetSingleton {
         Self { base, loader, saver }
     }
 }
+
+// Unregister the loader and saver when the extension exits.
```
Bromeon:
```suggestion
// Unregister the loader and saver when the extension is unloaded.
```
This makes a bit clearer that this also happens during extension reload.

Bromeon (commented):
Thank you!

src/recipes/resource-saver-loader.md
beicause: This is copied from https://godot-rust.github.io/book/recipes/engine-singleton.html#registering-a-singleton

I think it might fail because you could potentially unregister it elsewhere.

Bromeon: I see, thanks!

For the sake of this example, which is about resource savers/loaders, we don't need to be concerned about someone prematurely unregistering the singleton. Let's focus on the core logic to demonstrate 😉 ideally we'll provide a more automated way of registering singletons at some point anyway.

PR #108, opened 2025-08-20
Stringify function guidance

zshift:
Replaced a string literal with the `stringify!` macro, which is safer for refactoring/renames.

### DIFF
--- a/src/intro/hello-world.md
+++ b/src/intro/hello-world.md
@@ -449,7 +449,7 @@ impl Player {
     #[func]
     fn increase_speed(&mut self, amount: f64) {
         self.speed += amount;
-        self.base_mut().emit_signal("speed_increased", &[]);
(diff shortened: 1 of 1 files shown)


Bromeon:
Thanks for the contribution!

I don't see how this is more type-safe though. The argument of `stringify!` need not be a valid expression:
```rs
fn main() {
    println!("{}", stringify!(whatever));
}
```
prints `whatever`.

Instead, what we could do and what _is_ type-safe, is the typed signal API:
```rs
self.signals().speed_increased().emit();
```

zshift:
@Bromeon, I've updated it following your recommendation. Thanks for the response. I hadn't realized the `stringify!()` macro worked on arbitrary values.

Bromeon:
Thanks a lot!

A trick is sometimes to generate some code that accesses the identifier (so the compiler must check it), but is never executed at runtime. Can be as simple as `if false { ... }` or so.

PR #64, opened 2024-12-01
Update to examples that use Strings for passing arguments.

2-3-5-41:
tldr:
- Updated "Registering a singleton" example code.
- Changed all relevant examples of `string.into()` -> `&string`.
- Changed "MyEditorSingleton" -> "MyEngineSingleton" in the "engine-singleton.md" page.

PR #114, opened 2025-08-26
Correct ResourceFormatSaver example

Hakunamawatta:
I changed the `ResourceFormatSaver` to use the `recognize` method because `is_recognized_resource` is not a member of `IResourceFormatSaver`. I removed a false comment which claimed that it is necessary to check again whether the resource is recognized in `get_recognized_extensions`. When I tested my own saver, `get_recognized_extensions` was only called when `recognize` returned true.
I removed the `expect` call on the resource and replaced it with `is_some_and` because it is safer.

### DIFF
--- a/src/recipes/resource-saver-loader.md
+++ b/src/recipes/resource-saver-loader.md
@@ -130,21 +130,15 @@ impl IResourceFormatSaver for MyAssetSaver {
     ) -> PackedStringArray {
         let mut array = PackedStringArray::new();
         
-        // Even though the Godot docs state that you don't need this check, it is
-        // in fact necessary.
-        if Self::is_recognized_resource(res) {
-            // It is also possible to add multiple extensions per Saver.
-            array.push("myextension");
-        }
-        
+        // It is also possible to add multiple extensions per Saver.
+        array.push("myextension");
         array
     }
 
     // All resource types that this saver should handle must return true.
-    fn is_recognized_resource(res: Option<Gd<Resource>>) -> bool {
+    fn recognize(&self, res: Option<Gd<Resource>>) -> bool {
         // It is also possible to add multible resource types per Saver.
-        res.expect("Godot called this without an input resource?")
-            .is_class("MyResourceType")
+        res.is_some_and(|r| r.is_class("MyResourceType"))
     }
 
     // This defines your logic for actually saving your resource.


Bromeon:
Thanks! 

> because `is_recognized_resource` is not a member of `IResourceFormatSaver`.

Makes sense, thank you!

> I removed the `expect` call on the resource and replaced it with `is_some_and` because it is safer.

This changes semantics though -- previously an unexpected condition panicked (meaning a bug that should be fixed), while it now silently returns false. I'm not sure if I'd call this safer -- a precondition that the code relies on is no longer checked.

---

Hakunamawatta:
> The method is defined a few lines below, in the very same example.

It's not a member of `IResourceFormatSaver` and can't be defined there. That's why I replaced it with `recognize`, which is the actual method name. Also `is_recognized_resource` wasn't even a method; it was an associated function, which appears to be a mistake.

Bromeon:
I just noticed what you meant (edited in the very same second as your message :sweat_smile: ). I wonder if the original author meant to add it in a separate `impl` block, or in fact meant `recognize` -- but it seems your appraoch does the job, in a simpler way.

Hakunamawatta:
> a precondition that the code relies on is no longer checked.

This will be called when *any* resource is saved, not just custom ones. The purpose of this method is to check whether the saver can handle the `Resource` based on its type and panicking if the `Resource` is null defeats the purpose of the method. It's not that likely to happen though because `ResourceSaver` checks that the Resource is not null (you'd have to call the `ResourceFormatSaver` directly to provide a null resource). Most of Godot's implementations of `recognize` check for null.

kistz:
Hey :) 
im actually pretty confused by this myself ^^' .
My working code which i also based this example upon is in fact using recognize and also doesnt do the if check?

I have no idea or recollection of how this ended up in there maybe i batch renamed something? xD

Anyway `recognize` should definetly be the correct way thanks for catching it!

kistz:
Sry for another comment ^^

@Bromeon you actually added this in your first comit (#64) by accident i belive :) 

I dont know why you chose to do it that way but i do remember now that it used to be the case that even if `recognize` returned false the virtual method was still called maybe they fixed that bug in newer versions :)

Thats why i still put the check there in my first commit in 64 to confirm it is indeed the right loader.

If its confirmed working now without that check its good to go imo.

Have a great day and do ask if further questions arise!

Bromeon:
Thanks for responding after all this time!

> @Bromeon you actually added this in your first comit (#64) by accident i belive :)

Not sure what exactly you mean, there's no `recognize`, that's about singletons. Or the general pattern of `if` check before unregistering? Yes, that's right :slightly_smiling_face: but either way, good to hear we're on the right path!

kistz:
Damn thats a dumb typo ... i meant #65 the issue in which i added the chapter ^^

My first commit had the `recognize` function and your commit afterwards changed this to the current state :> 
I didnt double check so it slipped through possibly you wanted to give it a clearer name not realizing its a virtual function?

Doesnt matter either way thanks Hukuna for fixing this :)

Bromeon:
Oh now I see! Sorry for the confusion, and thanks to both of you! :muscle:

PR #115, opened 2025-09-03
docs(signals): add disclaimer about version requirement

Matthew-Burkard:
I was working with `godot = "0.2.3"`.
When copying from the book I was getting an error:

```
no method named `signals` found for mutable reference `&mut Monster` in the current scope method not found in `&mut Monster`
```

On any line attempting to call `self.signals()`. e.g. `let sig = self.signals().damage_taken();`

Fixed error bu switching from `godot = "0.2.3"` to `godot = { branch = "master", git = "https://github.com/godot-rust/gdext" }`

### DIFF
--- a/src/register/signals.md
+++ b/src/register/signals.md
@@ -130,6 +130,8 @@ For typed signals to be available, you need:
   - This must be an inherent impl, the `I*` trait `impl` won't be enough.
   - Leave the impl empty if necessary.
 - A `Base<T>` field.
+- The Latest version of the `godot` dependency.
+  - You may need to specify `godot = { branch = "master", git = "https://github.com/godot-rust/gdext" }` in your `Cargo.toml`.
 
 Signals, typed or not, **cannot** be declared in secondary `impl` blocks (those annotated with `#[godot_api(secondary)]` attribute).
 ```


Bromeon:
It's funny, I just removed this information [here](https://github.com/godot-rust/book/commit/10fdbba3fd20261aefc6804e0a4d4020145aac6e#diff-2fc1a182945c8d0260d050341463ed9967e39c484ae86575334caf2b13aa22d2). So it was useful after all :grinning:

Also, your change is not quite accurate -- you don't need master for this, [v0.3.0](https://github.com/godot-rust/gdext/blob/master/Changelog.md#v030) released in May supports typed signals.

So far, the book is aimed at the latest godot-rust version, sometimes even upcoming features. It can make sense to denote versions for new features or major ones that have been recently introduced (such as type-safe signals), but generally we need to find a balance to not micromanage entries. Any opinions on this?

Matthew-Burkard:
I should have checked for a newer version in crates.io first, was thinking I was on latest release for some reason.
Maybe a note saying 0.3.0+ but really for docs to describe latest release makes sense. I'm willing to chalk this up to me not realizing I was working with older release.

Yarwin:
> So far, the book is aimed at the latest godot-rust version, sometimes even upcoming features. It can make sense to denote versions for new features or major ones that have been recently introduced (such as type-safe signals), but generally we need to find a balance to not micromanage entries. Any opinions on this?

Hmm, book links to TypedSignal (<https://godot-rust.github.io/docs/gdext/master/godot/register/struct.TypedSignal.html>) which is available on current master but there is no convenient/fast way to check since when it has been available :thinking:. 

My idea – expand https://godot-rust.github.io/book/index.html#currently-supported-features and instead of linking to github issue, note that we encourage to use the latest version since all of them are backward-compatible, make an appendix in the book with currently supported features (table with feature - API docs - since when it is available) and link it there. 

As for the PR itself, your issue is valid, albeit this section <https://godot-rust.github.io/book/register/signals.html#admonition-availability-of-signal-api> is about availability of signal on given _class_, not the feature itself.

Bromeon:
> My idea – expand https://godot-rust.github.io/book/index.html#currently-supported-features and instead of linking to github issue, note that we encourage to use the latest version since all of them are backward-compatible, make an appendix in the book with currently supported features (table with feature - API docs - since when it is available) and link it there.

I think there's some value in having compatibility information on the page itself, as people are unlikely to navigate to this section.

Maybe we can keep it until 1 minor version back or so? Now that v0.4 is in development, we can mention v0.3 is the minimum, but once v0.4 is released, it would effectively be available in two minor versions and thus "recent enough" to no longer need a mention?

Because prior godot-rust versions are no longer maintained by us. But while they're not 100% compatible, we provide migration guides and changes are usually rather minor.

Bromeon:
Sorry I forgot about this, but as mentioned, by now typed signals are probably long enough in the library that we can assume general availability. Meanwhile v0.5 is in development -- if there are larger features getting their own new page, it can make sense to add such a compatibility note.

So I would close this, but thanks nonetheless for the contribution!

PR #117, opened 2025-09-20
Replace deprecated `dict!` macro with `vdict!`

JoseSK999:
I noticed the `dict!` macro used in the 3.1 section of the book is deprecated in favor of `vdict!`. I have just added the `v` letter in this commit.


Yarwin:
It would be good to update the description as well:

https://github.com/godot-rust/book/blob/7f16a41c4224d6a78803f126bc34cf72a26b6576/src/godot-api/builtins.md#L110-L111

Maybe something along the lines of

> [Dictionary](https://godot-rust.github.io/docs/gdext/master/godot/builtin/struct.Dictionary.html) is a key-value store, where both keys and values are Variant. Generic dictionaries are not currently supported in godot-rust.

?

JoseSK999:
So removing the `godot-generic-dicts` link?

Yarwin:
> So removing the `godot-generic-dicts` link?

Yep! No point in tracking this issue since this functionality has been merged into Godot already.

JoseSK999:
Updated!

PR #123, opened 2025-12-06
docs(debugger): Updated the debugger configuration for Visual Studio Code

greenpixels:
Updated the debugging documentation for VS Code as the prior configurations was outdated

These where the previous issues of the example configuration
- Warning: Missing property "program"
- The attempt to launch cross-platform seems to be not supported. You can't run platform per OS, only pass in arguments per OS to program via args https://code.visualstudio.com/docs/debugtest/debugging-configuration#_platformspecific-properties
- Going by the recommended project structure makes "preLaunchTask": "rust: cargo build" not execute correctly. Instead you need a separate tasks.json that builds rust in the correct folder. (See below)
With the given changes these should no longer be a problem.
- Removed the -e argument as starting the debug with godot in editor mode apparently does not keep any breakpoints set in Rust

I would have loved to omit the tasks.json, but sadly its not possible to set a working directory for `cargo build` ([except for running when nightly channel](https://doc.rust-lang.org/cargo/commands/cargo-build.html#common-options))

### DIFF
--- a/src/toolchain/debugging.md
+++ b/src/toolchain/debugging.md
@@ -17,32 +17,43 @@ you will only have symbols for stack frames in Rust code.
 
 ## Launching with VS Code
 
-Here is an example launch configuration for Visual Studio Code. Launch configurations should be added to  `./.vscode/launch.json`, relative
-to your project's root. This example assumes you have the [CodeLLDB] extension installed, which is common for Rust development.
+Here is an example debugger setup for Visual Studio Code. Launch configurations and tasks should be placed in`./.vscode/launch.json` and
+`./.vscode/tasks.json` respectively, relative to your project's root.
+
+This example assumes you have the [CodeLLDB] extension installed, which is common for Rust development.
 
 ```json
+// ./.vscode/launch.json
 {
     "configurations": [
         {
             "name": "Debug Project (Godot 4)",
             "type": "lldb", // type provided by CodeLLDB extension
             "request": "launch",
-            "preLaunchTask": "rust: cargo build",
-            "cwd": "${workspaceFolder}",
+            "cwd": "${workspaceFolder}/godot",
+            "preLaunchTask": "build-rust",
             "args": [
-                "-e", // run editor (remove this to launch the scene directly)
-                "-w", // windowed mode
+                "-w" // windowed mode
             ],
-            "linux": {
-                "program": "/usr/local/bin/godot4",
-            },
-            "windows": {
-                "program": "C:\\Program Files\\Godot\\Godot_v4.1.X.exe",
-            },
-            "osx": {
-                // NOTE: on macOS the Godot.app needs to be manually re-signed 
-                // to enable debugging (see below)
-                "program": "/Applications/Godot.app/Contents/MacOS/Godot",
+            "program": "PATH/TO/GODOT"
+        }
+    ]
+}
+```
+
+```json
+// ./.vscode/tasks.json
+{
+    "version": "2.0.0",
+    "tasks": [
+        {
+            "label": "build-rust",
+            "command": "cargo",
+            "args": [
+                "build"
+            ],
+            "options": {
+                "cwd": "${workspaceFolder}/rust"
             }
         }
     ]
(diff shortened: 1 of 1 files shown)


src/toolchain/debugging.md
```diff
@@ -17,32 +17,43 @@ you will only have symbols for stack frames in Rust code.
...
+This example assumes you have the [CodeLLDB] extension installed, which is common for Rust development.
+
+```jsonc
+// ./.vscode/launch.json
 {
     "configurations": [
         {
             "name": "Debug Project (Godot 4)",
-            "type": "lldb", // type provided by CodeLLDB extension
+            "type": "lldb",
```
Bromeon: Is the comment inaccurate, or why removed?

src/toolchain/debugging.md
```diff
@@ -17,32 +17,43 @@ you will only have symbols for stack frames in Rust code.
...
+            "args": [
+                "-w"
+            ],
+            "program": "PATH/TO/GODOT"
+        }
+    ]
+}
+```
+
+```jsonc
```
Bromeon: Note that while it's technically more correct, `jsonc` isn't associated with a syntax highlighter in mdbook, so the reader experience is worse than with `json`.

You can render the page with `mdbook serve --open` to quickly test this for yourself 🙂

src/toolchain/debugging.md
```diff
@@ -17,32 +17,43 @@ you will only have symbols for stack frames in Rust code.
...
-            "type": "lldb", // type provided by CodeLLDB extension
+            "type": "lldb",
             "request": "launch",
-            "preLaunchTask": "rust: cargo build",
-            "cwd": "${workspaceFolder}",
+            "cwd": "${workspaceFolder}/godot",
+            "preLaunchTask": "build-rust",
+            "args": [
+                "-w"
+            ],
```
Bromeon, line 40: Also here, you removed the `-e` and all comments. I understand why `-e` might not be preferrable as default choice, but you could still leave it there, commented-out, so that people who want it can quickly learn about it.

I don't see why you removed the comment for `-w`, it's not necessarily obvious what this argument does.

Bromeon (changes requested):
Thanks!

greenpixels:
Apologies, I removed the comments because it was for myself and I copied over my configuration promptly, forgetting correctly merging the two.

I will better consolidate it with the previous version.

src/toolchain/debugging.md
greenpixels: Opening it as editor with -e does not keep the breakpoints for LLDB. At least I was unable to make it work and I found several other sharing that problem on the discord.

I forgot mentioning this in my original PR description and will add it now.

src/toolchain/debugging.md
greenpixels: See https://github.com/godot-rust/book/pull/123#issuecomment-3621375650

src/toolchain/debugging.md
greenpixels: Yea, I quickly edited it in GitHubs Web Editor 🙃 I'll do it proper and run it locally, sorry for the holdup.

src/toolchain/debugging.md
Yarwin: Middleground: just link https://docs.godotengine.org/en/latest/tutorials/editor/command_line_tutorial.html#command-line-reference.

Yarwin:
Note – I had issue with lldb in which I had to manually add `br s -r rust_panic` to make it stop on rust panics, not only the breakpoints – is it a case with VSCode?

In Zed I'm handling it by specifying: `"initCommands": ["br s -r rust_panic"],` in my config.

-----

Oh, and remember about squashing commits later :sweat_smile:

greenpixels:
Changed to json instead of jsonc and added back comments.
Unable to build locally due to mdbook issues I've mentioned on the discord (migration issues)

I also did a local squash rebase, but that seems to not have worked here ... my local git history looks good now, but seems it didn't transfer correctly.

Anyhow, if the PR is still not satisfactory then please treat it as stale or close it as I'm currently short on time, sorry!

PR #124, opened 2025-12-21
Add info about `ConnectHandle`.

Yarwin:


src/register/signals.md
```diff
@@ -240,6 +240,39 @@ impl INode3D for Monster {
...
+        self.handle = Some(self.signals()
+            .damage_taken()
+            .connect(Self::on_damage_taken));
+    }
+}
+
+impl Monster {
+    fn make_invulnerable(&mut self) {
+        if let Some(connection) = self.handle.take()
+        && connection.is_connected()
```
Bromeon:
```suggestion
            && connection.is_connected()
```
(Rustfmt would indent this)

But I would not recommend `is_connected()` as general practice; it's needed only if there is no clear ownership of the signal, or if the involved object is expected to die. Maybe there could be a comment about a potential panic instead.

src/register/signals.md
```diff
@@ -240,6 +240,39 @@ impl INode3D for Monster {
 ```
 
 
+## Disconnecting
+
+Connecting a signal in any aforementioned way returns a `ConnectHandle`, which can be used for disconnecting.
```
Bromeon: Maybe add a link to API docs?

Bromeon (commented):
Great example with invulnerability 👍

src/register/signals.md
```diff
@@ -240,6 +240,85 @@ impl INode3D for Monster {
...
+struct Commentator {
+    handle: Option<ConnectHandle>,
+    base: Base<Node3D>,
+}
+
+impl Commentator {
+    fn watch_monster(&mut self, monster: Gd<Monster>) {
+        if let Some(connection) = self.handle.take()
+            && connection.is_connected()
+        {
```
Bromeon, line 289:
```suggestion
        if let Some(connection) = self.handle.take()
            /* && connection.is_connected() -- see above */
        {
```

Bromeon: Or even just 
```rs
if let Some(connection) = self.handle.take() { 
```

Yarwin: I did it on purpose this time, see comment over this section:

> Disconnecting a signal will result in panic if given connection does not exist.
> Use `is_connected()` when there is no clear ownership of the signal or connected object is expected to be freed.

alternatively I can remove the whole 2nd section :thinking:.

Bromeon (approved):

Bromeon:
Thank you!

PR #127, opened 2026-01-05
Web export: add `-Zdefault-visibility=hidden`

Bromeon:
See https://github.com/godot-rust/gdext/issues/968#issuecomment-3712268752

### DIFF
--- a/src/toolchain/export-web.md
+++ b/src/toolchain/export-web.md
@@ -52,8 +52,8 @@ This is platform-specific.
 
 ## Project Configuration
 
-Prebuilts for wasm32 are not provided – consult [Custom Godot versions][custom-godot-version]
-to configure your project with [`api-custom`][api-cargo-features] feature.
+At the moment, you'll need to use [`api-custom`][api-cargo-features] to export to the web. In the future, we will provide a "prebuilt"
+target (like Linux, macOS and Windows) directly for Wasm32. For now, see [Custom Godot versions][custom-godot-version].
 
 Enable the [`experimental-wasm`][api-cargo-features] feature on godot-rust in the `Cargo.toml` file.
 It is also recommended to enable the [`lazy-function-tables`][api-cargo-features] feature to avoid long compile times with release builds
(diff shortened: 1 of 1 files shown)


SvizelPritula (approved):
Looks good!

src/toolchain/export-web.md
```diff
@@ -84,12 +84,23 @@ rustflags = [
     "-C", "link-args=-pthread", # /!\ Read 'Thread support' below regarding this flag
     "-C", "target-feature=+atomics", # /!\ Read 'Thread support' below regarding this flag
     "-C", "link-args=-sSIDE_MODULE=2",
+    "-C", "-Zdefault-visibility=hidden",
```
PgBiel: I think you meant this?

```suggestion
    "-Z", "default-visibility=hidden",
```

src/toolchain/export-web.md
```diff
@@ -240,6 +252,7 @@ Here's how this can be done:
         RUSTFLAGS="-C link-args=-pthread \
         -C target-feature=+atomics \
         -C link-args=-sSIDE_MODULE=2 \
+        -C -Zdefault-visibility=hidden \
```
PgBiel:
```suggestion
        -Z default-visibility=hidden \
```

src/toolchain/export-web.md
```diff
@@ -456,6 +469,7 @@ RUSTFLAGS="-C link-args=-g \
 -C link-args=-pthread \
 -C target-feature=+atomics \
 -C link-args=-sSIDE_MODULE=2 \
+-C default-visibility=hidden \
```
PgBiel:
```suggestion
-Z default-visibility=hidden \
```

src/toolchain/export-web.md
```diff
@@ -84,12 +84,23 @@ rustflags = [
...
     "-Zlink-native-libraries=no",
     "-Cllvm-args=-enable-emscripten-cxx-exceptions=0",
     "-Z", "emscripten-wasm-eh=false",
 ]
 ```
 
+A bit of background: The compiled Wasm binary is a _side module_ which is dynamically loaded by Godot's _main module_ at runtime.
+This is very similar to how dynamic libraries and executables work.
+
+The [`default-visibility=hidden` flag][rustc-visibility] ensures that symbols of each side module are private to that module.
```
PgBiel, line 97: Could this affect user (non-gdext internal) code somehow? If this is the default, how can it be selectively turned off?

src/toolchain/export-web.md
PgBiel, line 420: Wonder if we could also mention something about this bug in this section, but I suppose it doesn't lead to one specific error, rather being undefined behavior / a race condition.

src/toolchain/export-web.md
Bromeon: I currently don't see a use case where one wouldn't want that -- it would mean explicitly sharing symbols with other side modules (Godot or extensions) -- and there are better ways to communicate with them (GDExtension itself) :thinking:

Bromeon:
Thanks a lot @PgBiel! 

The different formatting and order of `-C` and `-Z` totally confused me -- I made this consistent now.

Regarding "what if one forgets `-Zdefault-visibility=hidden`", I thought about this yesterday. An idea would be to install a canary in godot-rust, e.g. verifying on startup that a global variable is unpopulated, and crashing with a useful error otherwise.

src/toolchain/export-web.md
SvizelPritula: I don't think it can be selectively turned off right now, which is unfortunate. There is [an RFC](https://github.com/rust-lang/rfcs/pull/3834), though.

SvizelPritula: I still think that the toolchain's behaviour is buggy, I believe normally only `#[unsafe(no_mangle)]` should be exported like this. As I mentioned in my comment, on x86-64 Linux all non-mangled objects end up as `LOCAL`.

PgBiel:
> An idea would be to install a canary in godot-rust, e.g. verifying on startup that a global variable is unpopulated, and crashing with a useful error otherwise.

Interesting. Would be nice indeed i possible.

More generally, though, I really wish we could just manage those flags for the user. Would be nice if we could figure out a way to set "defaults"...

I remember looking into some solution involving `build.rs`, but I don't remember how far that went. If possible, though, this could become part of some project template or similar.

PR #136, opened 2026-03-03
Update Singleton page to current function naming

pbiswell:
Hi,

This just updates the 'Engine singletons' page to have the current function/enum/class names.
It also matches the registering code with the struct name that's in the previous section.

Thanks,
Paul

### DIFF
--- a/src/recipes/engine-singleton.md
+++ b/src/recipes/engine-singleton.md
@@ -83,7 +83,7 @@ unsafe impl ExtensionLibrary for MyExtension {
     fn on_stage_deinit(stage: InitStage) {
         if stage == InitStage::Scene {
             // Let's keep a variable of our Engine singleton instance,
-            // and MyEngineSingleton name.
+            // and MyEditorSingleton name.
             let mut engine = Engine::singleton();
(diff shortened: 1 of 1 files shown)


Bromeon:
Thanks!

I think `MyEditorSingleton` is misleading, it suggests that it would only run in the editor like `#[class(tool)]`.
The previous section wasn't good in that regard either...

Could you change all occurrences to simply `MySingleton`? Including others on that page.

pbiswell:
I've updated those now.

Thanks,
Paul

Yarwin:
Note that we support user-defined singletons: https://godot-rust.github.io/docs/gdext/master/godot/prelude/derive.GodotClass.html#user-defined-singletons.

Bromeon:
Yes, I think we should extend [that page](https://godot-rust.github.io/book/recipes/engine-singleton.html?highlight=MyEditorSingleton#defining-a-singleton) with `#[class(singleton)]`, as added by Yarwin in https://github.com/godot-rust/gdext/pull/1399, maybe even starting the page with that.

But this can be done separately -- the improvements in this PR are valuable either way. Even with proc-macro singletons, there are situations where manually registering singletons can be needed.

PR #137, opened 2026-03-08
Add info about hot reload signals interaction.

Yarwin:
Follow up to: https://github.com/godot-rust/gdext/pull/1512

### DIFF
--- a/src/register/signals.md
+++ b/src/register/signals.md
@@ -126,7 +126,7 @@ The signal type is implementation-defined. Besides the `#[signal]`-specific cust
 ```admonish note title="Availability of signal API"
 For typed signals to be available, you need:
 
-- A `#[godot_api] impl MyClass {}` block.  
+- A `#[godot_api] impl MyClass {}` block.
   - This must be an inherent impl, the `I*` trait `impl` won't be enough.
   - Leave the impl empty if necessary.
 - A `Base<T>` field.
@@ -200,7 +200,7 @@ impl INode3D for Monster {
         // Let's say damage is deflected to a shield object.
         // That one is stored as field `shield: OnReady<Gd<Shield>>`.
         // &*self.shield is thus the `&Gd<Shield>` we need.
-        
+
         self.signals()
             .damage_taken()
             .connect_other(&*self.shield, Shield::on_damage_taken);
@@ -228,7 +228,7 @@ impl INode3D for Monster {
         self.signals()
             .damage_taken()
             .connect(Self::on_damage_taken);
-        
+
         // Or with closures:
         self.signals()
             .damage_taken()
@@ -392,7 +392,7 @@ impl INode3D for Monster {
         self.signals()
             .damage_taken()
             .connect_other(&*self.shield, Shield::on_damage_taken);
-        
+
         // Connect to the `Node::renamed` signal, which is invoked
         // when a node name changes.
         self.signals()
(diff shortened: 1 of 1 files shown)


src/register/signals.md
```diff
@@ -556,6 +556,48 @@ Just use closures instead. Generally speaking, the [`TypedSignal`][api-typedsign
 to be extensible for your own workflows.
 
 
+### Signals in the editor – hot reload interaction
+
+Typed signals use Custom Callables for connections. Custom Callables which contain a vtable – a set of function pointers required
+to manage their functionality. After hot reload all these function pointers become invalid and invoking any of them results
+in undefined behavior, usually leading to crash.
```
Bromeon:
```suggestion
Typed signals use custom callables for connections. Custom callables which contain a vtable – a set of function pointers pointing
to Godot FFI functions. After hot reload, all these function pointers become invalid, and invoking any of them results
in undefined behavior, usually leading to crash.
```

Bromeon: Also, maybe the first occurrence of "callable" in the text could point to the API docs for `Callable` in general?

src/register/signals.md
```diff
@@ -556,6 +556,48 @@ Just use closures instead. Generally speaking, the [`TypedSignal`][api-typedsign
...
+
+Non-custom Callables are ones created with [`Callable::from_object_method`][api-callable-from-object-method] or [`Gd::callable()`][api-gd-callable].
```
Bromeon: Note that this has technically been an implementation detail so far. By documenting which constructors use custom callables internally, we restrict how those are implemented. For example, `from_class_static` is currently implemented as custom callable, but if GDExtension provides a native constructor for static methods, this may change.

A safer (and official) way to check which callables are custom would be to query the `Callable::is_custom()` method -- maybe this could be mentioned first, and then something like "at the moment, `from_object_method` + ... can be used to create standard (non-custom) callables."

We might also want to link here from API docs.

src/register/signals.md
```diff
@@ -556,6 +556,48 @@ Just use closures instead. Generally speaking, the [`TypedSignal`][api-typedsign
...
+let non_custom_callable: Callable = monster.callable("damage_taken");
+// Will not cause any problems upon hot reload.
+monster.connect("damage_taken", &non_custom_callable);
+
+let custom_callable: Callable = Callable::from_fn(
+    "my custom callable", 
+    |_| godot_print!("hello world!")
+);
+// Invoking  this signal after hot reload is unsound and will lead to crash!
+monster.connect("damage_taken", &custom_callable);
```
Bromeon:
```suggestion

// Invoking this signal after hot reload will lead to UB.
monster.connect("damage_taken", &custom_callable);
```

(also empty line)

src/register/signals.md
```diff
@@ -556,6 +556,48 @@ Just use closures instead. Generally speaking, the [`TypedSignal`][api-typedsign
...
+to manage their functionality. After hot reload all these function pointers become invalid and invoking any of them results
+in undefined behavior, usually leading to crash.
+
+Non-custom Callables are ones created with [`Callable::from_object_method`][api-callable-from-object-method] or [`Gd::callable()`][api-gd-callable].
+
+```rs
+let monster: Gd<InEditorMonster> = ...;
+let non_custom_callable: Callable = monster.callable("damage_taken");
+// Will not cause any problems upon hot reload.
+monster.connect("damage_taken", &non_custom_callable);
```
Bromeon: Please leave empty lines before comments, to easily recognize grouped lines.

```suggestion

// Will not cause any problems upon hot reload.
monster.connect("damage_taken", &non_custom_callable);
```

src/register/signals.md
```diff
@@ -556,6 +556,48 @@ Just use closures instead. Generally speaking, the [`TypedSignal`][api-typedsign
...
+let custom_callable: Callable = Callable::from_fn(
+    "my custom callable", 
+    |_| godot_print!("hello world!")
+);
+// Invoking  this signal after hot reload is unsound and will lead to crash!
+monster.connect("damage_taken", &custom_callable);
+```
+
+To prevent unsoundness, godot-rust automatically disconnects all typed signal connections (i.e. created using anything other
+than the untyped signals API) before a hot reload.
```
Bromeon, line 566: Just a thought: connections that are created through the untyped signal API, but involve custom callables, are not covered -- meaning those still suffer from the UB?

If yes, that's probably worth mentioning in a footnote. It's a very rare use case, but worth documenting for completeness.

src/register/signals.md
```diff
@@ -556,6 +556,48 @@ Just use closures instead. Generally speaking, the [`TypedSignal`][api-typedsign
...
+These connections can be recreated afterwards, typically using the [`EXTENSION_RELOADED`][api-extension-reloaded-notification] notification.
+Alternatively, non-typed signals – which are preserved during hot reload – with non-custom Callables can be used.
```
Bromeon:
```suggestion
Alternatively, untyped signals – which are preserved during hot reload – with standard callables can be used.
```

src/register/signals.md
```diff
@@ -556,6 +556,48 @@ Just use closures instead. Generally speaking, the [`TypedSignal`][api-typedsign
...
+
+```rs
+#[godot_api]
+impl INode for InEditorMonster {
+    fn on_notification(&mut self, what: NodeNotification) {
+        if what == NodeNotification::EXTENSION_RELOADED {
+            let objects = Vec::from(&self.managed_editor_objects);
```
Bromeon: It's not clear what `managed_editor_objects` is (never defined), nor why a `Vec` is created here.

I guess `Vec` is because of `self` borrowing in later `signals()` call -- but this should be mentioned in a short comment.

Bromeon (commented):
I'd use the term "standard" instead of "non-custom" to avoid constant negation. It's also the [term that Godot uses](https://docs.godotengine.org/en/stable/classes/class_callable.html#class-callable-method-is-standard).

src/register/signals.md
Yarwin: hmm, example was supposed to illustrate that - but mentioning it explicitly won't hurt.

src/register/signals.md
Yarwin: good idea, but `Callable::is_custom` is misleading in this context. Godot considers a product of `Callable(...).bind(...)` to be a custom callable.

https://docs.godotengine.org/en/stable/classes/class_callable.html#class-callable-method-is-custom

> Returns true if this Callable is a custom callable. Custom callables are used:
> 
> - for binding/unbinding arguments (see [bind()](https://docs.godotengine.org/en/stable/classes/class_callable.html#class-callable-method-bind) and [unbind()](https://docs.godotengine.org/en/stable/classes/class_callable.html#class-callable-method-unbind));
> 
> - for representing methods of built-in [Variant](https://docs.godotengine.org/en/stable/classes/class_variant.html#class-variant) types (see [create()](https://docs.godotengine.org/en/stable/classes/class_callable.html#class-callable-method-create));
> 
> -  for representing global, lambda, and RPC functions in GDScript;
> 
> - for other purposes in the core, GDExtension, and C#.

We care only about the very last case – everything else should be sound.

I'll figure out something, hold on

src/register/signals.md
Yarwin: I've finally decided to use `.clone()`, but I added necessary context.

Partial borrows :weary:.

src/register/signals.md
```diff
@@ -556,13 +556,70 @@ Just use closures instead. Generally speaking, the [`TypedSignal`][api-typedsign
 to be extensible for your own workflows.
 
 
+### Signals in the editor – hot reload interaction
+
+
```
Bromeon: Extra empty line... Strange that the linter doesn't detect this 🤔

src/register/signals.md
```diff
@@ -556,13 +556,70 @@ Just use closures instead. Generally speaking, the [`TypedSignal`][api-typedsign
...
+// Will not cause any problems upon hot reload.
+monster.connect("damage_taken", &non_custom_callable);
+
+let custom_callable: Callable = Callable::from_fn(
+    "my custom callable",
+    |_| godot_print!("hello world!")
+);
+
+// Invoking  this signal after hot reload is unsound and will lead to UB!
```
Bromeon:
```suggestion
// Invoking this signal after hot reload will lead to UB!
```

(either ! or . is fine)

Bromeon (approved):

PR #138, opened 2026-03-08
Update custom engine singletons page.

Yarwin:


src/recipes/engine-singleton.md
```diff
@@ -5,7 +5,7 @@
   ~ file, You can obtain one at https://mozilla.org/MPL/2.0/.
 -->
 
-# Engine singletons
+# User engine singletons
```
Bromeon:
```suggestion
# User singletons
```
I know the term "engine singleton" was used to differentiate from "autoloads", but imo:
- it's enough if we use "singleton" consistently across all docs
- there is already a box on that at the very top
- for everything else we use "engine" as the counterpart to "user", so it would be good to stay consistent with _that_ terminology

Applies to all mentions of the term "engine singleton" on this page; I don't highlight the others.

src/recipes/engine-singleton.md
```diff
@@ -39,104 +37,140 @@ Godot provides _many_ built-in singletons in its API. You can find a full list [
...
+## Using a singleton with `on_main_loop_frame`
+
+Since Godot4.5+ [`on_main_loop_frame`][api-main-loop-frame] can be used to invoke an user singleton:
+
+```rs
+fn global_delta() -> f64 {
+    1. / ProjectSettings::singleton()
+        .get("physics/common/physics_ticks_per_second")
+        .to::<i64>() as f64
+}
```
Bromeon, line 76: Please store the fetched value in a variable, _then_ do the calculation.


```suggestion
fn global_delta() -> f64 {
    let ticks = ProjectSettings::singleton()
        .get("physics/common/physics_ticks_per_second")
        .to::<i64>();
        
    1.0 / (ticks as f64)
}
```

src/recipes/engine-singleton.md
```diff
@@ -39,104 +37,140 @@ Godot provides _many_ built-in singletons in its API. You can find a full list [
...
+#[class(init, singleton)]
+struct MySingleton {
+    #[init(val = global_delta())]
+    delta: f64,
+    #[init(val = true)]
+    paused: bool,
+    #[init(val = Instant::now())]
+    time: Instant,
+    base: Base<Object>,
+}
```
Bromeon, line 91:
```suggestion
struct MySingleton {
    #[init(val = global_delta())]
    delta: f64,

    #[init(val = true)]
    paused: bool,

    #[init(val = Instant::now())]
    time: Instant,

    base: Base<Object>,
```

src/recipes/engine-singleton.md
```diff
@@ -39,104 +37,140 @@ Godot provides _many_ built-in singletons in its API. You can find a full list [
...
+There should be only one instance of a given singleton class in the engine, valid as long as the library is loaded.
+Therefore, user singletons are limited to classes with manual memory management (ones not inheriting from RefCounted).
 
-func _ready() -> void:
-    MySingleton.foo()
-```
+```rs
+#[derive(GodotClass)]
+#[class(init, base = Object)]
+struct MyEngineSingleton {}
```
Bromeon: Please keep `MySingleton`, was just renamed in #136; see also comment about naming above.

src/recipes/engine-singleton.md
```diff
@@ -39,104 +37,140 @@ Godot provides _many_ built-in singletons in its API. You can find a full list [
...
-[refcounted-singleton-issue]: https://github.com/godot-rust/gdext/issues/522
 
+## Registering custom engine singleton without proc macro
 
-## Calling from GDScript
+Custom engine singleton can be registered through [`godot::classes::Engine`][api-class-engine].
+Additionally, implementing [`UserSingleton`][api-user-singleton] allows accessing a registered singleton instance through `singleton()`.
 
-Now that your singleton is available (and once you've recompiled and reloaded), you should be able to access it from GDScript like so:
+User singletons should be registered under their class name – otherwise some Godot components (for example GDScript before 4.4)
```
Bromeon, line 130: I generally use "--" rather than "–", easier to type for most people. AFAIR they're rendered the same in mdbook (or rustdoc?). Would also help us potentially detect AI slop easier 😉

Bromeon (commented):
Thanks a lot! Great content -- just some remarks about naming

src/recipes/engine-singleton.md
Yarwin: I started with `--` and sadly mdbook doesn't render them as – >:[

Bromeon (approved):

PR #153, opened 2026-09-19
Add gdext_async crate

Mercerenies:
I have published a very small crate that acts as a shim layer between Godot's `await` and Rust's `await` (`gdext-coroutines` looks neat but my crate is lighter weight and for people who want to stay on stable Rust).

This PR adds that crate to the ecosystem page.


Bromeon (approved):
Nice, thanks a lot for contributing to the ecosystem 🚀 

Over time you might want to add a motivating example in the repo's ReadMe, but I saw you just got started. All the best for the project 😊

